PR #1, opened 2021-05-28
Universe - added methods for testing - + added SourceData class for testing

lindajohnstone:
Hardened Universe testing

### DIFF
--- a/ConwaysGameOfLife.Tests/SourceData.cs
+++ b/ConwaysGameOfLife.Tests/SourceData.cs
@@ -1,0 +1,16 @@
+namespace ConwaysGameOfLife.Tests
+{
+    public static class SourceData
+    {
+        public static Universe UniverseInitializedAllDeadCells()
+        {
+            var initData = new CellState[][]
+            {
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead },
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead },
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead }
+            };
+            return new Universe(initData);
+        }
+    }
+}
--- a/ConwaysGameOfLife/Universe.cs
+++ b/ConwaysGameOfLife/Universe.cs
@@ -15,9 +15,46 @@ namespace ConwaysGameOfLife
 
         public Universe(int gridWidth, int gridLength)
         {
+            Initialize(gridWidth, gridLength, InitializeCells(gridWidth, gridLength));
+        }
+
+        public Universe(CellState[][] sourceData)
+        {
+            Initialize(sourceData.GetLength(0), sourceData.Length, InitializeCells(sourceData));
+        }
+        public Universe(Universe sourceUniverse)
+        {
+            Initialize(sourceUniverse.GridWidth, sourceUniverse.GridLength, sourceUniverse._cells);
+        }
+        
+        public void Initialize(int gridWidth, int gridLength, List<Cell> cells)
+        {
             GridWidth = gridWidth;
             GridLength = gridLength;
-            _cells = InitializeCells(gridWidth, gridLength);
+            _cells = cells;
+        }
+        
+        private List<Cell> InitializeCells(CellState[][] sourceData)
+        {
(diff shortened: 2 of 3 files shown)


ConwaysGameOfLife.Tests/SourceData.cs
```diff
@@ -0,0 +1,16 @@
...
+            var initData = new CellState[][]
+            {
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead },
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead },
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead }
+            };
+            return new Universe(initData);
+        }
+    }
+}
```
kedoodle, line 16:
```suggestion
}

```
add newline to end of file - this is just for you to see a review suggestion and how you can commit it directly to the branch from github's UI

ConwaysGameOfLife.Tests/SourceData.cs
```diff
@@ -0,0 +1,16 @@
...
+    public static class SourceData
+    {
+        public static Universe UniverseInitializedAllDeadCells()
+        {
+            var initData = new CellState[][]
+            {
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead },
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead },
+                new []  {CellState.Dead,   CellState.Dead,    CellState.Dead }
+            };
```
kedoodle, line 12: could we consider other options instead of a 2D array of cellstate, given that for this kata a List is used to store cells instead of a 2D array

you could have a helper function in here to create a list of cells which are passed to Universe in the constructor. we can talk about this option on monday//use it as a refactoring exercise

eg. input:
```csharp
"OOO" + Environment.NewLine +
XXX"
```

eg. output:
```csharp
var output = List<Cell> {
  new Cell(0, 0, CellState.Dead),
  new Cell(1, 0, CellState.Dead),
  new Cell(2, 0, CellState.Dead),
  new Cell(0, 1, CellState.Alive),
  new Cell(1, 1, CellState.Alive),
  new Cell(2, 1, CellState.Alive),
  new Cell(0, 2, CellState.Dead),
  new Cell(1, 2, CellState.Dead),
  new Cell(2, 2, CellState.Dead)
}
```

ConwaysGameOfLife/Universe.cs
```diff
@@ -14,10 +14,47 @@ public class Universe
...
+                {
+                    if (sourceData[y].Length != gridLength)
+                    {
+                        throw new ArgumentOutOfRangeException("sourceData should be a square array");
+                    }
+                    _cells.Add(new Cell(sourceData[x][y], x, y));
+                }
+            }
+            return _cells;
         }
```
indianbread, line 58: a lot of this logic could live in a helper method in the sourcedata. ideally, we should minimise the amount of code in our actual classes that is used just for testing purposes

ConwaysGameOfLife/Universe.cs
```diff
@@ -14,10 +14,47 @@ public class Universe
...
+                for (var y = 0; y < gridLength; y++)
+                {
+                    if (sourceData[y].Length != gridLength)
+                    {
+                        throw new ArgumentOutOfRangeException("sourceData should be a square array");
+                    }
+                    _cells.Add(new Cell(sourceData[x][y], x, y));
+                }
+            }
+            return _cells;
```
kedoodle, line 57: can we also please double check the x and y values when indexing a 2D array? i think your outer loop should be looping the y/length, and the inner loop indexing the x/width

kedoodle (approved):
well done 👍 , a few things we can discuss // opportunities to refactor next week

ConwaysGameOfLife.Tests/SourceData.cs
```diff
@@ -0,0 +1,16 @@
...
+            return new Universe(initData);
+        }
+    }
```
indianbread, line 15: perhaps the source data can be structured a different way so you don't have to create a copy constructor in the actual universe class just for the purposes of testing

indianbread (approved):

PR #2, opened 2021-05-31
Moved SwitchCellState method to Cell. Attempt to write helper method

lindajohnstone:


ConwaysGameOfLife.Tests/CellShould.cs
```diff
@@ -54,5 +54,15 @@ public void WhenCellStateAndLocationAreSame_BeEqual()
...
+        [Fact]
+        public void WhenUniverseRegenerated_SwitchState()
+        {
+            var cell = new Cell(CellState.Dead, 0, 0);
+            var expected = new Cell(CellState.Alive, 0, 0);
+
+            var result = cell.SwitchCellState(cell);
+
+            Assert.Equal(expected, result);
+        }
```
Kedoodle, line 66: Do we want the Cell to know how to regenerate itself? Or is it "dumb" and some higher class is responsible for regenerating the Universe as a whole, and thereby the Cells?

ConwaysGameOfLife.Tests/TestUniverse.cs
```diff
@@ -0,0 +1,51 @@
...
+                    {
+                        throw new ArgumentOutOfRangeException("sourceData should be a square array"); // TODO: ?? need exception? - if yes, change message
+                    }
+                    // split into characters
+                    var currentCell = src[y][x];
+                    _cells.Add(new Cell(GetSourceDataCellState(currentCell.ToString()), x, y));
+                }
+            }
+            return new Universe(gridWidth, gridLength);
+        }
```
Kedoodle, line 36: good start on this refactor in moving the helper method here and changing the sourceData to String type.

we can work on the logic tomorrow. potentially need a constructor in the Universe which just takes a list of cells?

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -128,5 +118,17 @@ public void WhenInitialized_FindCellStateOfCell_UsingSourceData(CellState state,
 
             Assert.Equal(expected, result);
         }
+
+        [Fact]
+        public void testName() // test being used to test method 
+        {
+            var initData = "XXX" + Environment.NewLine + "XXX" + Environment.NewLine + "XXX";
```
kedoodle, line 125:
```suggestion
            var initData = "XXX" + Environment.NewLine +
                                   "XXX" + Environment.NewLine +
                                   "XXX";
```

you may find it easier to interpret like this

ConwaysGameOfLife.Tests/CellShould.cs
kedoodle, line 66: Do we want the Cell to know how to regenerate itself? Or is it "dumb" and some higher class is responsible for regenerating the Universe as a whole, and thereby the Cells?

ConwaysGameOfLife.Tests/TestUniverse.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                return CellState.Dead;
+            }
+            return CellState.Alive;
+        }
+        private static string[] SplitInput(string input, string delimiter)
+        {
+            return input.Split(delimiter, StringSplitOptions.RemoveEmptyEntries);
+        }
+    }
```
kedoodle, line 50: good start on this refactor in moving the helper method here and changing the sourceData to String type.

we can work on the logic tomorrow. potentially need a constructor in the Universe which just takes a list of cells?

ConwaysGameOfLife/Cell.cs
```diff
@@ -16,6 +16,15 @@ public Cell(CellState state, int x, int y)
...
 
+        public Cell SwitchCellState(Cell cell) // changes the state of the cell
+        {
+            if (cell.IsAlive())
+            {
+                return new Cell(CellState.Dead, cell.Location.X, cell.Location.Y);
+            }
+            return new Cell(CellState.Alive, cell.Location.X, cell.Location.Y);
+        }
+
```
kedoodle, line 27: same as first comment:

> Do we want the Cell to know how to regenerate itself? Or is it "dumb" and some higher class is responsible for regenerating the Universe as a whole, and thereby the Cells?

kedoodle (approved):
nice work getting started, lets crack on tomorrow

ConwaysGameOfLife.Tests/CellShould.cs
indianbread: @Kedoodle this was my doing . I interpreted this as toggling the state of the cell so asked Linda to move the method here.
@lindajohnstone  your options:
1. If the cell can toggle its state, keep the method here
2. If the universe is responsible for regenerating itself as a whole and thereby the cells, rename the test to better reflect that a new cell is created with a different state and move this method back to the universe. To paraphrase uncle Bob, when naming a test, it does what it says and says what it does.

kedoodle: no worries. the current name implies the universe being regenerated, but there is no universe instance or regeneration of that instance.

PR #3, opened 2021-06-01
21 06 01

lindajohnstone:


ConwaysGameOfLife/Universe.cs
```diff
@@ -7,9 +7,9 @@ namespace ConwaysGameOfLife
     public class Universe
     {
         // creates the Universe/World
-        private List<Cell> _cells;
+        public List<Cell> Cells { get; private set; }
         private List<Location> _neighbourLocations;
```
kedoodle, line 11: would like to pick your brains about this at some point

kedoodle (approved):
awesome stuff today. let's continue with the equality stuff tomorrow.

spoiler: we don't necessarily need the custom equality stuff for our "business logic", only for testing purposes. we're doing it as a learning exercise, after which we can consider using libraries like nhan alluded to.

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+            if (!isSizeSame)
+            {
+                return false;
+            }
+            // test location using new method - for loop using index
+            for (var x = 0; x < obj1.GridLength; x++)
+            {
+                for (var y = 0; y < obj1.GridLength; y++)
+                {
+                    if (obj1.GetCellAtLocation(x, y) != obj2.GetCellAtLocation(x, y))
```
indianbread, line 17: this is a good start, adding a note here as a reminder to refactor this tomorrow. Options:
1. Add another helper method to check if 2 cells are equivalent
2. Do a manual check if cell state 

Also thought of a better name for this function - UniversesAreEquivalent

PR #4, opened 2021-06-02
21 06 02

lindajohnstone:


ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -25,5 +30,71 @@ public static bool UniverseContentsAreEqual(Universe obj1, Universe obj2)
...
+            var isLocationSame = (obj1.Location.X == obj2.Location.X) && (obj1.Location.Y == obj2.Location.Y);
+            if (!isLocationSame)
+            {
+                return false;
+            }
+            if (obj1.CellState != obj2.CellState)
+            {
+                return false;
+            }
+            return true;
```
kedoodle, line 45: I like the `isLocationSame` intermediate variable. We could do a similar thing for the cell state and then combine the return statements.
```suggestion
            var isLocationSame = (obj1.Location.X == obj2.Location.X) && (obj1.Location.Y == obj2.Location.Y);
            var isCellStateSame = obj1.CellState == obj2.CellState;
            return isLocationSame && isCellStateSame;
```

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -1,22 +1,27 @@
...
-            // test location using new method - for loop using index
+            
             for (var x = 0; x < obj1.GridLength; x++)
             {
                 for (var y = 0; y < obj1.GridLength; y++)
                 {
-                    if (obj1.GetCellAtLocation(x, y) != obj2.GetCellAtLocation(x, y))
+                    var cell1 = obj1.GetCellAtLocation(x, y);
+                    var cell2 = obj2.GetCellAtLocation(x, y);
+                    if (cell1 != cell2)
```
kedoodle, line 22: Is this check needed? I'd imagine it will always be false unless you pass `UniversesAreEqual` the same `obj1` and `obj2`

ConwaysGameOfLife.Tests/CellShould.cs
```diff
@@ -38,18 +38,16 @@ public void WhenInitialized_SetLocation(int x, int y)
             
             Assert.Equal(new Location(x, y), result);
         }
-        [Fact]
-        public void WhenCellStateAndLocationAreSame_BeEqual()
+
+        [Theory]
+        [InlineData(0, 1, CellState.Dead, 0, 1, CellState.Dead, true)]
+        [InlineData(0, 1, CellState.Alive, 0, 1, CellState.Dead, false)]
+        public void WhenCellStateAndLocationAreSame_BeEqual(int x1, int y1, CellState oneState, int x2, int y2, CellState twoState, bool expected) // TODO: rename method
```
kedoodle, line 45: Let's tidy up the naming to be consistent (using both `one`/`two` and `1`/`2`). Additionally, the order of the params could be consistent with the signature of the `Cell` constructor:

```suggestion
        [Theory]
        [InlineData(CellState.Dead, 0, 1, CellState.Dead, 0, 1, true)]
        [InlineData(CellState.Alive, 0, 1, CellState.Dead, 0, 1, false)]
        public void WhenCellStateAndLocationAreSame_BeEqual(CellState state1, int x1, int y1, CellState state2, int x2, int y2, bool expected) // TODO: rename method
```

ConwaysGameOfLife.Tests/CellShould.cs
```diff
@@ -38,18 +38,16 @@ public void WhenInitialized_SetLocation(int x, int y)
...
-            var y2 = 1;
-            var one = new Cell(CellState.Dead, x1, y1);
-            var two = new Cell(CellState.Dead, x2, y2);
-            var expected = true;
-
-            var result = (one.Equals(two));
+            var one = new Cell(oneState, x1, y1);
+            var two = new Cell(twoState, x2, y2);
+
+            var result = UniverseHelper.CellsAreEqual(one, two);
```
kedoodle, line 50: Same as above

```suggestion
            var cell1 = new Cell(state1, x1, y1);
            var cell2 = new Cell(state2, x2, y2);

            var result = UniverseHelper.CellsAreEqual(cell1, cell2);

```

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -25,5 +30,71 @@ public static bool UniverseContentsAreEqual(Universe obj1, Universe obj2)
...
+            {
+                return false;
+            }
+            for (var cell = 0; cell < obj1.Count; cell++)
+            {
+                if (!CellsAreEqual(obj1[cell], obj2[cell]))
+                {
+                    return false;
+                }
+            }
```
kedoodle, line 60: What happens if the cells in the list are in a different order?

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -36,19 +36,21 @@ public void WhenUniverseGenerated_ReturnsListOfNeighbours()
...
+                new Location(1, 0),
+                new Location(1, 2),
+                new Location(2, 0),
+                new Location(2, 1),
+                new Location(2, 2),
+            };
 
             var result = universe.GetCellNeighbourLocations(cell);
 
             Assert.Equal(expected, result);
```
kedoodle, line 52: Is this reliant on an override?

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -85,17 +87,17 @@ public void WhenInitialized_ReturnsEmptyListOfAliveCells()
         }
 
         [Fact]
-        public void testName() // test being used to test method 
+        public void testName() // TODO: rename
         {
             var initData = "XXX" + Environment.NewLine + 
                                     "XXX" + Environment.NewLine + 
                                     "XXX";
```
kedoodle, line 94: Can we align the rows so that it's easier to see? I think the original suggestion on github messed up the alignment
image

ConwaysGameOfLife/Universe.cs
```diff
@@ -67,23 +65,25 @@ public List<Location> GetCellNeighbourLocations(Cell cell)
...
+            return cellAtLocation.CellState;
         }
 
         public List<Cell> CheckIfNeighboursAlive(Cell cell) 
         {
             var neighbourList = new List<Cell>();
-            _neighbourLocations = GetCellNeighbourLocations(cell);
-            foreach (var neighbour in _neighbourLocations)
+            var neighbourLocations = GetCellNeighbourLocations(cell);
+            foreach (var neighbour in neighbourLocations)
```
kedoodle, line 81: Could potentially use Linq:

```suggestion
            aliveNeighbours = neighbourLocations.Count(location => GetCellStateFromLocation(location) == CellState.Alive);
```

ConwaysGameOfLife/Universe.cs
```diff
@@ -67,23 +65,25 @@ public List<Location> GetCellNeighbourLocations(Cell cell)
...
-            var newCell = newCells.ElementAt(0);
-            return newCell.CellState;
+            // var newCell = Cells.Find(cell => cell.Location == new Location(cell.Location.X, cell.Location.Y));
+            var x = location.X;
+            var y = location.Y;
+            var cellAtLocation = Cells.FirstOrDefault(cell => cell.Location.X == x && cell.Location.Y == y);
+            return cellAtLocation.CellState;
         }
 
         public List<Cell> CheckIfNeighboursAlive(Cell cell)
```
kedoodle, line 77: The name implies that we want to know if the neighbours are alive
- do we need to return a list of cells? is that what we're expecting? should it return a boolean instead?
- is it all neighbours are alive?
- is it some neighbours are alive?
- is it one neighbour is alive?

What do we need this function for? Probably because later on, we want to implement rules that are based on the number of alive neighbours.
```suggestion
        public int CountAliveNeighbours(Cell cell)
```

ConwaysGameOfLife/Universe.cs
```diff
@@ -67,23 +65,25 @@ public List<Location> GetCellNeighbourLocations(Cell cell)
...
+            foreach (var neighbour in neighbourLocations)
             {
                 var state = GetCellStateFromLocation(neighbour);
                 if (state == CellState.Alive)
                 {
-                    _neighbourLocations.Add(neighbour);
+                    neighbourLocations.Add(neighbour);
                 }
             }
             return neighbourList;
```
kedoodle, line 89: As it stands, `neighbourList` will always return an empty list

ConwaysGameOfLife/Universe.cs
```diff
@@ -67,23 +65,25 @@ public List<Location> GetCellNeighbourLocations(Cell cell)
...
-            _neighbourLocations = GetCellNeighbourLocations(cell);
-            foreach (var neighbour in _neighbourLocations)
+            var neighbourLocations = GetCellNeighbourLocations(cell);
+            foreach (var neighbour in neighbourLocations)
             {
                 var state = GetCellStateFromLocation(neighbour);
                 if (state == CellState.Alive)
                 {
-                    _neighbourLocations.Add(neighbour);
+                    neighbourLocations.Add(neighbour);
```
kedoodle, line 86: adding to the wrong place?

```suggestion
                    neighbourList.Add(neighbour);
```

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -124,5 +126,17 @@ public void WhenGivenALocation_ReturnsCell()
...
+        public void TestUniverseHelper_UniversesAreEqual() // TODO: rename method
+        {
+            var gridWidth = 4;
+            var gridLength = 4;
+            var universe = new Universe(gridWidth, gridLength);
+
+            var result = UniverseHelper.UniversesAreEqual(universe, new Universe(4, 4));
+
+            Assert.True(result);
+        }
```
kedoodle, line 140: Probably worth instantiating the two universes in different ways e.g. one as on line 135 and another with a List of cells similar to the neighbour tests.

kedoodle (approved):
good stuff with the refactors today. a whopping 121 lines deleted with 136 added!

i've added a few comments for us to think about/discuss

some of the comments in the `Universe` `CheckIfNeighboursAlive` method should be read together

ConwaysGameOfLife.Tests/CellShould.cs
indianbread: sorry late to the party but can also look at using **member data** to allow you to use objects for inline data

PR #5, opened 2021-06-03
21 06 03

lindajohnstone:


ConwaysGameOfLife/Universe.cs
```diff
@@ -53,40 +53,33 @@ public bool AreAllCellsDead()
...
+                new Location(cell.Location.X - 1, cell.Location.Y + 1),     //0,2
+                new Location(cell.Location.X, cell.Location.Y - 1),         //1,0
+                new Location(cell.Location.X, cell.Location.Y + 1),         //1,2
+                new Location(cell.Location.X + 1, cell.Location.Y - 1),     //2,0
+                new Location(cell.Location.X + 1, cell.Location.Y),         //2,1
+                new Location(cell.Location.X + 1, cell.Location.Y + 1)      //2,2
+            };
+            
             return neighbourLocations;
         }
```
kedoodle, line 68: note: we still need to account for neighbours of cells that touch the universe boundary/boundaries

ConwaysGameOfLife/Neighbours.cs
```diff
@@ -18,29 +18,32 @@ public Neighbours(Cell cell)
...
+                new Cell(cell.CellState, cell.Location.X - 1, cell.Location.Y + 1),     //0,2
+                new Cell(cell.CellState, cell.Location.X, cell.Location.Y - 1),         //1,0
+                new Cell(cell.CellState, cell.Location.X, cell.Location.Y + 1),         //1,2
+                new Cell(cell.CellState, cell.Location.X + 1, cell.Location.Y - 1),     //2,0
+                new Cell(cell.CellState, cell.Location.X + 1, cell.Location.Y),         //2,1
+                new Cell(cell.CellState, cell.Location.X + 1, cell.Location.Y + 1)      //2,2
+            };
             // loop through each location
             // use location (how?) to determine cell state
             return CellNeighbours;
```
kedoodle, line 49: The `Neighbours` class will have trouble with finding neighbour locations/cells in the context of an actual universe as it's not "universe aware". Consider what this class is intended to do:
- is it supposed to be able to find the neighbours?
- or is it just a "struct" to store the neighbours of a cell? does it need to exist at all? (i have no preference as to whether it exists or not, just wanting to provoke some discussion here)

ConwaysGameOfLife/Universe.cs
```diff
@@ -53,40 +53,33 @@ public bool AreAllCellsDead()
...
-
-        public List<Cell> CheckIfNeighboursAlive(Cell cell) 
-        {
-            var neighbourList = new List<Cell>();
+            var aliveNeighbours = new List<Location>();
             var neighbourLocations = GetCellNeighbourLocations(cell);
             foreach (var neighbour in neighbourLocations)
             {
-                var state = GetCellStateFromLocation(neighbour);
+                var state = GetCellAtLocation(neighbour.X, neighbour.Y).CellState;
```
kedoodle, line 76: "reaching in" here. should separate into a few lines and do a null check i.e. assign the cell to a variable, check the cell isn't null, (potentially assign the state of the cell to a variable) check the state

kedoodle (approved):
looking good.

in general, we probably want to do more test cases 
for boolean return types, we want to test both the `true` and `false` cases
e.g. if we're testing that a function considers two universes equal, we should also write test cases for when they shouldn't be equal. otherwise, how do we know that the function doesn't just always consider two universes equal? it could just be `return true` and the tests would still pass.
in general, we might want to test boundary conditions where we expect logic to be different in some way
e.g. if we're testing a function that count alive neighbour cells, we might want to test a "normal" cell that isn't on the boundary, a cell that is on one boundary (x), a cell that is on the other boundary (y), and a cell that is on both boundaries (x and y i.e. in a corner).

PR #6, opened 2021-06-04
21 06 04

lindajohnstone:


ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -33,17 +33,29 @@ public void WhenXAndYAreSame_BeEqualOperator()
             Assert.True(result);
         }
 
-        [Fact]
-        public void WhenTwoLocationsInstantiated_BeEqual()
+        [Theory]
+        [InlineData(0, 0, 0, 0, true)]
+        [InlineData(0, 0, 1, 0, false)]
+        public void WhenTwoLocationsInstantiated_BeEqual(int x1, int y1, int x2, int y2, bool expected)
```
kedoodle, line 39: probably should only have true cases here( and then Assert.True later)
```suggestion
        [InlineData(0, 0, 0, 0)]
        [InlineData(2, 2, 2, 2)]
        public void WhenTwoLocationsInstantiated_BeEqual(int x1, int y1, int x2, int y2)

```

ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -33,17 +33,29 @@ public void WhenXAndYAreSame_BeEqualOperator()
...
+        [Theory]
+        [InlineData(0, 0, 1, 0, false)]
+        public void WhenTwoLocationsInstantiated_NotBeEqual(int x1, int y1, int x2, int y2, bool expected)
+        {
+            var location1 = new Location(x1, y1);
+            var location2 = new Location(x2, y2);
+
+            var result = UniverseHelper.LocationsAreEqual(location1, location2);
+
+            Assert.Equal(expected, result);
```
kedoodle, line 58: same as above but false
```suggestion
        [Theory]
        [InlineData(0, 0, 1, 0)]
        [InlineData(1, 2, 2, 2)]
        public void WhenTwoLocationsInstantiated_NotBeEqual(int x1, int y1, int x2, int y2)
        {
            var location1 = new Location(x1, y1);
            var location2 = new Location(x2, y2);

            var result = UniverseHelper.LocationsAreEqual(location1, location2);

            Assert.False(result);

```

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -61,11 +61,11 @@ public static bool ListsOfCellsAreEqual(List<Cell> obj1, List<Cell> obj2)
 
         public static bool LocationsAreEqual(Location obj1, Location obj2)
         {
-            if (obj1.X != obj2.X && obj1.Y != obj2.Y)
+            if (obj1.X == obj2.X && obj1.Y == obj2.Y)
```
kedoodle, line 64: good job removing the double negative.

could replace obj1 and obj2 with better names now that we're using custom helpers instead of overrides (can no longer just pass in any old object). this goes for all of the equality helpers

```suggestion
            if (location1.X == location2.X && location1.Y == location2.Y)

```

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -144,5 +132,31 @@ public void WhenTwoUniversesInstantiated_BeEqual()
 
             Assert.True(result);
         }
+
+        [Fact]
+        public void WhenTwoUniversesInstantiated_NotBeEqual()
```
kedoodle, line 137:
```suggestion
        public void WhenUniversesOfDifferentDimensionsInstantiated_NotBeEqual()

```

ConwaysGameOfLife/Universe.cs
```diff
@@ -82,6 +78,14 @@ public int CountLiveNeighbours(Cell cell)
             return aliveNeighbours.Count;
         }
 
+        private static void ThrowsExceptionIfCellIsNull(Cell cell)
+        {
+            if (cell == null)
+            {
+                throw new NullReferenceException();
+            }
```
kedoodle, line 86: have a look into null coalescing if you're interested

```suggestion
            cell ?? throw new NullReferenceException();

```

kedoodle (approved):
good to see more tests coming in. we can look into adding more functionality next week! have a great weekend.

.gitignore
```diff
@@ -91,6 +91,8 @@ fabric.properties
 !.vscode/launch.json
 !.vscode/extensions.json
 *.code-workspace
+launch.json
```
indianbread, line 94: don't include this in .gitignore as your application usually uses this to read environment variables. e.g. for an application that's deployed, this is where you would store your environment variable so the application knows the environment is development and not production when you run it on your local machine

ConwaysGameOfLife.Tests/LocationShould.cs
indianbread: Agree with Keddy. Try to only test 1 thing per test, one reason being so that if something fails you can easily identify which one. I recommend using multiple asserts in 1 test only if they are related (e.g. when comparing the contents of 2 cells and needed to have 3 asserts to check that it's the same)

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -144,5 +132,31 @@ public void WhenTwoUniversesInstantiated_BeEqual()
...
+
+        [Fact]
+        public void WhenCellIsNull_ThrowException() // TODO: rename - if cell in GetCellNeighbourLocations is null & throws exception
+        {
+            Cell cell = null;
+            var universe = new Universe(3, 3);
+
+            var result = Assert.Throws<NullReferenceException>(() => universe.GetCellNeighbourLocations(cell));
+            
+            Assert.Equal("Object reference not set to an instance of an object.", result.Message);
```
indianbread, line 159: My opinion: I don't think you need to test the error message specifically as they can be customised/changed, so you don't want to have to update your test if you decide to modify your error message. It's enough to test that the null reference error will be thrown if the cell is null. Also when asserting that an error is thrown, you don't have to assign this to a variable.

indianbread (approved):

PR #7, opened 2021-06-07
21 06 07

lindajohnstone:


ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -57,5 +29,61 @@ public void WhenTwoLocationsInstantiated_NotBeEqual(int x1, int y1, int x2, int
...
+                new Location(2, 0),
+                new Location(2, 1),
+                new Location(2, 2)
+            };
+
+            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
+        }
+
+        [Fact]
+        public void WhenTwoListsOfCellsNotSameOrderInstantiated_BeEqual()
```
indianbread, line 62: the naming of this should be list of locations as above

ConwaysGameOfLife.Tests/RulesShould.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using Moq;
+using Xunit;
+
+namespace ConwaysGameOfLife.Tests
+{
+    public class RulesShould
+    {
+        [Fact]
+        public void WhenAnyLiveCellHasFewerThan2LiveNeighbours_ItDies()
```
indianbread, line 10: keen to hear your thought process behind how this will work

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -33,35 +33,35 @@ public static bool UniversesAreEqual(Universe obj1, Universe obj2)
...
         {
-            var isLocationSame = (obj1.Location.X == obj2.Location.X) && (obj1.Location.Y == obj2.Location.Y);
-            var isCellStateSame = obj1.CellState == obj2.CellState;
+            var isLocationSame = (cell1.Location.X == cell2.Location.X) && (cell1.Location.Y == cell2.Location.Y);
+            var isCellStateSame = cell1.CellState == cell2.CellState;
             return isCellStateSame && isLocationSame;
         }
 
-        public static bool ListsOfCellsAreEqual(List<Cell> obj1, List<Cell> obj2)
+        public static bool ListsOfLocationsAreEqual(List<Location> location1, List<Location> location2)
```
indianbread, line 43: suggestion: use plural for a variable associated with a collection (list)

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -33,35 +33,35 @@ public static bool UniversesAreEqual(Universe obj1, Universe obj2)
...
             {
                 return false;
             }
-            var obj1Ordered = obj1.OrderBy(c => c.Location.X).ThenBy(c => c.Location.Y).ToList();
-            var obj2Ordered = obj2.OrderBy(c => c.Location.X).ThenBy(c => c.Location.Y).ToList();
+            var location1Ordered = location1.OrderBy(c => c.X).ThenBy(c => c.Y).ToList();
+            var location2Ordered = location2.OrderBy(c => c.X).ThenBy(c => c.Y).ToList();
 
-            for (var cell = 0; cell < obj1.Count; cell++)
+            for (var location = 0; location < location1.Count; location++)
```
indianbread, line 52: suggestion: rename location to i as it's an index and not a location

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -20,7 +20,7 @@ public void WhenInitialized_BePopulatedWithAllDeadCells()
         }  
 
         [Fact]
-        public void WhenUniverseGenerated_ReturnsListOfNeighbours()
+        public void WhenGivenANonBoundaryCell_ReturnCorrectNeighbours()
```
indianbread, line 23: nice naming

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -134,7 +134,7 @@ public void WhenTwoUniversesInstantiated_BeEqual()
         }
 
         [Fact]
-        public void WhenTwoUniversesInstantiated_NotBeEqual()
+        public void WhenUniversesOfDifferentDimensionsInstantiated_NotBeEqual()
```
indianbread, line 137: the test is not doing what the name is suggesting.

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -154,9 +154,113 @@ public void WhenTwoUniversesInstantiated_NotBeEqual()
...
+                new Location(1, 2)
+            };
+
+            var result = universe.GetCellNeighbourLocations(cell);
+
+            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
+        }
+
+        [Fact]
+        public void WhenGivenABoundaryCellX_ReturnCorrectNeighbourLocationsListOutOfOrder()
```
indianbread, line 204: what is the purpose of this test compared to the previous one? Is it required?

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -154,9 +154,113 @@ public void WhenTwoUniversesInstantiated_NotBeEqual()
...
+                new Location(1, 2)
+            };
+
+            var result = universe.GetCellNeighbourLocations(cell);
+
+            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
+        }
+
+        [Fact]
+        public void WhenGivenABoundaryCellY_ReturnCorrectNeighbourLocationsListOutOfOrder()
```
indianbread, line 225: does the order of the locations matter, as long as they're correct?

ConwaysGameOfLife/Universe.cs
```diff
@@ -48,15 +48,21 @@ public bool AreAllCellsDead()
 
         public List<Location> GetCellNeighbourLocations(Cell cell)
         {
+            var xMiddle = cell.Location.X;
```
indianbread, line 51: suggestion: maybe call this plain x and y

ConwaysGameOfLife/Universe.cs
```diff
@@ -65,7 +71,7 @@ public List<Location> GetCellNeighbourLocations(Cell cell)
         public int CountLiveNeighbours(Cell cell)
         {
             var aliveNeighbours = new List<Location>();
```
indianbread, line 73: alternative: `var aliveNeighbours = 0;`
then do the check in the loop and increment by 1 if the cell is alive

indianbread (approved):
good work setting up the neighbours & checking the boundary conditions!

ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -1,41 +1,13 @@
...
-            var one = location;
-            var two = new Location(0, 0);
-
-            var result = (one == two);
-
-            Assert.True(result);
-        }
-
         [Theory]
         [InlineData(0, 0, 0, 0, true)]
```
kedoodle, line 10: Can we add an extra case here, remove the `expected` param, and `Assert.True` instead
```suggestion
        [InlineData(0, 0, 0, 0)]
        [InlineData(5, 5, 5, 5)]
```

kedoodle: Same goes for the `WhenTwoLocationsInstantiated_NotBeEqual` test, but `Assert.False`

ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -57,5 +29,61 @@ public void WhenTwoLocationsInstantiated_NotBeEqual(int x1, int y1, int x2, int
 
             Assert.Equal(expected, result);
         }
+
+        [Fact]
+        public void WhenTwoListsOfLocationsSameOrderInstantiated_BeEqual()
```
kedoodle, line 34: No need to say `Instantiated` as we're not really testing the constructor for a class you've written
```suggestion
        public void GivenTwoListsOfLocationsSameOrder_BeEqual()
```

ConwaysGameOfLife.Tests/LocationShould.cs
kedoodle, line 62: Same goes re: `Instantiated`
```suggestion
        public void GivenTwoListsOfLocationsNotSameOrder_BeEqual()
```

ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -1,41 +1,13 @@
...
-
-            var result = (one == two);
-
-            Assert.True(result);
-        }
-
         [Theory]
         [InlineData(0, 0, 0, 0, true)]
-        [InlineData(0, 0, 1, 0, false)]
         public void WhenTwoLocationsInstantiated_BeEqual(int x1, int y1, int x2, int y2, bool expected)
```
kedoodle, line 11: No need to say Instantiated as we're not really testing the constructor for a class you've written

```suggestion
        public void GivenTwoLocationsOfSameCoordinates_BeEqual(int x1, int y1, int x2, int y2)
```

kedoodle: Same goes for the `WhenTwoLocationsInstantiated_NotBeEqual `test  i.e. could be renamed to `GivenTwoLocationsOfDifferingCoordinates_NotBeEqual`

ConwaysGameOfLife.Tests/RulesShould.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace ConwaysGameOfLife.Tests
+{
+    public class RulesShould
+    {
+        [Fact]
+        public void WhenAnyLiveCellHasFewerThan2LiveNeighbours_ItDies()
+        {
+            var universe = new Universe(3, 3);
+            var rules = new Rules(universe);
+            var cell = new Cell(It.IsAny<CellState>(), 1, 1);
```
kedoodle, line 14: Do we need a `Cell` or just a `Location` here? What is the purpose of stubbing the `CellState`?

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -6,25 +6,25 @@ namespace ConwaysGameOfLife.Tests
 {
     public static class UniverseHelper
     {
-        public static bool UniversesAreEqual(Universe obj1, Universe obj2)
+        public static bool UniversesAreEqual(Universe universe1, Universe universe2)
         {
-            var isSizeSame = (obj1.GridWidth == obj2.GridWidth) && (obj1.GridLength == obj2.GridLength);
+            var isSizeSame = (universe1.GridWidth == universe2.GridWidth) && (universe1.GridLength == universe2.GridLength);
```
kedoodle, line 11: Extra consideration: this is actually only checking the grid dimensions - what about if the actual number of `Cells` were to be different? Could be an extra check early on which is just as simple as this existing one

ConwaysGameOfLife.Tests/UniverseHelper.cs
kedoodle: either `locations1` or `locationList1` works

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -33,35 +33,35 @@ public static bool UniversesAreEqual(Universe obj1, Universe obj2)
...
                 }
             }
             return true;
         }
 
-        public static bool LocationsAreEqual(Location obj1, Location obj2)
+        public static bool LocationsAreEqual(Location location1, Location location2)
         {
-            if (obj1.X == obj2.X && obj1.Y == obj2.Y)
+            if (location1.X == location2.X && location1.Y == location2.Y)
```
kedoodle, line 64: can just return this like 
```suggestion
            return location1.X == location2.X && location1.Y == location2.Y;
```

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -37,7 +37,7 @@ public void WhenUniverseGenerated_ReturnsListOfNeighbours()
 
             var result = universe.GetCellNeighbourLocations(cell);
 
-            Assert.Equal(expected, result);
+            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
             Assert.Equal(8, result.Count);
```
kedoodle, line 41: the count equality should be encapsulated in the first assertion
```suggestion
            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
```

ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -57,5 +29,61 @@ public void WhenTwoLocationsInstantiated_NotBeEqual(int x1, int y1, int x2, int
...
+                new Location(1, 0),
+                new Location(2, 0),
+                new Location(2, 1),
+                new Location(2, 2)
+            };
+
+            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
+        }
     }
 }
```
kedoodle, line 89:
```suggestion
}

```

ConwaysGameOfLife.Tests/RulesShould.cs
```diff
@@ -0,0 +1,21 @@
...
+
+            var result = rules.Check(cell);
+
+            Assert.True(result);
+        }
+    }
+}
```
kedoodle, line 21:
```suggestion
}

```

ConwaysGameOfLife.Tests/UniverseShould.cs
kedoodle: could be a sign that the helper method isn't doing as it should i.e.
- shouldn't throw exceptions
- should always return true or false

ConwaysGameOfLife.Tests/UniverseShould.cs
kedoodle: Sorry I may have suggested that it's okay to have the expected list in a different order without clarifying that the `UniverseHelper.ListsOfLocationsAreEqual` should already handle when the lists aren't in the same order

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -154,9 +154,113 @@ public void WhenTwoUniversesInstantiated_NotBeEqual()
...
+                new Location(1, 1),
+                new Location(1, 2)
+            };
+
+            var result = universe.GetCellNeighbourLocations(cell);
+
+            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
         }
     }
 }
```
kedoodle, line 266:
```suggestion
}

```

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -154,9 +154,113 @@ public void WhenTwoUniversesInstantiated_NotBeEqual()
...
+                new Location(1, 2)
+            };
+
+            var result = universe.GetCellNeighbourLocations(cell);
+
+            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
+        }
+
+        [Fact]
+        public void WhenGivenABoundaryCornerCell_ReturnCorrectNeighbourLocationsListOutOfOrder()
```
kedoodle, line 246: love that you're testing several "kinds" of locations i.e.
- non boundary
- single boundary
- dual boundary

could think about extending this with `[Theory]` to become:
- non boundary
- single boundary on y axis (case each for top and bottom)
- single boundary on x axis (case each for left and right)
- all four corners?

ConwaysGameOfLife/Location.cs
```diff
@@ -15,52 +15,5 @@ public Location(int x, int y)
...
-        public static bool operator !=(Location obj1, Location obj2)
-        {
-            if (!OperatorOverride(obj1, obj2))
-            {
-                return false;
-            }
-            return true;
-        }
     }
 }
```
kedoodle, line 19:
```suggestion
}

```

ConwaysGameOfLife/OutputFormatter.cs
```diff
@@ -24,7 +24,7 @@ public string FormatWelcomeMessage()
             return "Welcome to Game of Life!";
         }
 
-        public string FormatUniverse()
+        public string FormatUniverse() // universe created with user input - dimensions & live cells ?? - needed??
```
kedoodle, line 27: we can consider writing some tests for this formatter even though we haven't started actually outputting anything to the console

```suggestion
        public string FormatUniverse(Universe universe) // universe created with user input - dimensions & live cells ?? - needed??
```

ConwaysGameOfLife/Rules.cs
```diff
@@ -8,6 +10,13 @@ public class Rules
         // * Any live cell with more than three live neighbours dies, as if by overcrowding.
         // * Any live cell with two or three live neighbours lives on to the next generation.
         // * Any dead cell with exactly three live neighbours becomes a live cell.
-        
+        public Rules(Universe universe)
```
kedoodle, line 13: Which way should this dependency be? We're currently injecting a `Universe` into an instance of `Rules`

ConwaysGameOfLife/Universe.cs
kedoodle: we paired on this, I mentioned `xMiddle` could be more explicit than `x` if you want to avoid potential confusion. I have no preference personally

ConwaysGameOfLife/Universe.cs
kedoodle: can also rejig this function (with some more widespread refactoring) to do a Linq count. we can discuss this tomorrow

ConwaysGameOfLife/Universe.cs
```diff
@@ -65,7 +71,7 @@ public List<Location> GetCellNeighbourLocations(Cell cell)
         public int CountLiveNeighbours(Cell cell)
         {
             var aliveNeighbours = new List<Location>();
-            ThrowsExceptionIfCellIsNull(cell);
```
kedoodle, line 68: I'm not a fan of throwing random unhandled exceptions. There are more graceful ways of handling stuff like this e.g. simply not passing any potential nulls to `CountLiveNeighbours`

ConwaysGameOfLife/Rules.cs
```diff
@@ -8,6 +10,13 @@ public class Rules
...
+        {
+        }
+
+        public bool Check(Cell cell)
+        {
+            throw new NotImplementedException();
+        }
     }
 }
```
kedoodle, line 22:
```suggestion
}

```

kedoodle (approved):
a few nitpicks and improvements. nice work on getting the neighbour tests and function going, as well as starting on the rules.

there's an opportunity to do a bunch of refactoring once the neighbour stuff is finalised - have a think about what calls will actually be made on the `Universe` eventually (hint: this could even be none! or it could be a fraction of what public functions are currently available)

PR #10, opened 2021-06-10
21 06 10

lindajohnstone:


ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -0,0 +1,25 @@
...
+            _generator = new Generator(new Universe(3, 3));
+        }
+        [Fact]
+        public void WhenInitialized_DisplayUniverse()
+        {
+            var output = new StubOutput();
+            var formatter = new OutputFormatter();
+            var expected = ". . . \n. . . \n. . . \n";
+
+            _generator.DisplayUniverse(output, formatter);
```
kedoodle, line 20: What is the responsibility of the Generator? Should it be displaying the Universe that it generates?

ConwaysGameOfLife.Tests/InputParserShould.cs
```diff
@@ -6,65 +6,69 @@ namespace ConwaysGameOfLife.Tests
...
+        }
         [Theory]
         [InlineData("0,0", 0, 0)]
         [InlineData("-1,-1", -1, -1)]
         [InlineData("100,8", 100, 8)]
-        public void WhenUserEntersStringOfOneAliveCellLocation_ReturnsListOfLocationContainingOnlyOneValue(string userInput, int x, int y) //TODO: shorten method name
+        [InlineData("a,0", -1, 0)]
+        [InlineData("p,s", -1, -1)]
+        [InlineData("3,p", 3, -1)]
+        [InlineData("*,2", -1, 2)]
```
kedoodle, line 21: I'm not sure how to approach this @indianbread any suggestions? Potentially a `TryParse` function which returns true/false for whether the parse was successful and a Location output into a variable when it is successful.

ConwaysGameOfLife.Tests/LocationShould.cs
```diff
@@ -20,7 +22,7 @@ public class LocationShould // TODO: all tests relate to universehelper methods
 
         [Theory]
         [InlineData(0, 0, 1, 0)]
```
kedoodle, line 24: can we add another test case or two here?

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -66,7 +44,7 @@ public void WhenInitialized_FindCellStateOfCell(CellState state, int x, int y)
         [InlineData("XXX", "OXX", "XXX", 1, 1, 1 )]
         [InlineData("OOO", "OOO", "OOO", 1, 1, 8)]
         [InlineData("OXX", "XXX", "XXO", 1, 1, 2)]
```
kedoodle, line 46: Given the deletion of tests for neighbour locations, we should add more tests cases here to cover the correct counting boundary cells. Potentially split into `ReturnCountOfLiveNeighbours_GivenBoundaryCoordinates` and `ReturnCountOfLiveNeighbours_GivenNonBoundaryCoordinates`

ConwaysGameOfLife/InputParser.cs
```diff
@@ -19,6 +19,18 @@ public List<Location> ParseListOfLocationsFromUserInput(string userInput)
...
+        public Universe ParseUniverse(string userInput)
+        {
+            var universe = SplitInput(userInput, ",");
+            return new Universe(ParseInput(universe[0]), ParseInput(universe[1]));
+        }
+
+        public Location ParseLocation(string input)
+        {
+            var location = SplitInput(input, ",");
+            return new Location(ParseInput(location[0]), ParseInput(location[1]));
```
kedoodle, line 31: should we be validating before or after using the input? i.e. do we want to attempt to create the universe if the input isn't valid?

ConwaysGameOfLife/Output/IOutput.cs
```diff
@@ -3,7 +3,8 @@ namespace ConwaysGameOfLife
     public interface IOutput
     {
         // leave open for extension - other output instead of console output - write file (save game?)
-        public void ConsoleWriteLine(string value);
-        public void ConsoleWrite(string value);
+        public void Write(string value);
+
+        public void WriteLine(string value);
```
kedoodle, line 8: good

kedoodle (approved):
good stuff refactoring and getting started with the validator. we might need to think about the flow to get from user input -> object created. we don't really want to be creating invalid objects, so how can we ensure that we have valid inputs before they're passed to the constructors for our objects?

ConwaysGameOfLife.Tests/InputParserShould.cs
indianbread: Agree @Kedoodle

ConwaysGameOfLife.Tests/ValidatorShould.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    public class ValidatorShould
+    {
+        Validator _validator;
+        public ValidatorShould()
+        {
+            _validator = new Validator();
+        }
+        [Fact]
+        public void WhenValidParsedInput_IsUniverse()
```
indianbread, line 14:
```suggestion
public void AcceptUniverse_WhenInputIsValid()
```

ConwaysGameOfLife/Generator.cs
```diff
@@ -23,6 +25,23 @@ rules checked
...
+
+        
+
+        public Generator(Universe universe)
+        {
+            _universe = universe;
+            
+        }
+
+        public void DisplayUniverse(IOutput output, OutputFormatter formatter)
```
indianbread, line 41: FYI C# has abstractions of Console.In and Console.Out called [Textwriter ](https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter?view=net-5.0)and [Textreader](https://docs.microsoft.com/en-us/dotnet/api/system.io.textreader?view=net-5.0) which saves you from having to create your own interface for them

PR #11, opened 2021-06-11
21 06 11

lindajohnstone:


ConwaysGameOfLife.Tests/InputParserShould.cs
```diff
@@ -6,69 +6,45 @@ namespace ConwaysGameOfLife.Tests
...
 
-            var result = _parser.ParseLocation(input);
+            var result = InputParser.ParseLocation(input);
 
             Assert.True(UniverseHelper.LocationsAreEqual(expected, result));
         }
 
         [Fact]
-        public void ReturnLocations_GivenAStringOfLocations()
+        public void NotReturnNewLocation_FromInvalidString()
```
indianbread, line 23:
```suggestion
        public void NotReturnNewLocation_GivenInvalidString()

```
For consistency with the rest of your test names

ConwaysGameOfLife.Tests/InputParserShould.cs
```diff
@@ -6,69 +6,45 @@ namespace ConwaysGameOfLife.Tests
...
         [Theory]
         [InlineData("3,4", 3, 4)]
-        [InlineData("a,4", -1, 4)]
-        [InlineData("10,a", 10, -1)]
-        [InlineData("a,0", -1, 0)]
-        [InlineData("p,s", -1, -1)]
-        [InlineData("3,p", 3, -1)]
-        [InlineData("*,2", -1, 2)]
+        [InlineData("4,4", 4, 4)]
         public void ReturnNewUniverse_FromString(string input, int x, int y)
```
indianbread, line 33:
```suggestion
        public void ReturnNewUniverse_GivenValidString(string input, int x, int y)

```

ConwaysGameOfLife/InputParser.cs
```diff
@@ -3,32 +3,32 @@
...
-                liveCellLocations.Add(new Location(ParseInput(location[0]), ParseInput(location[1])));
-            }
-            return liveCellLocations;
-        }
 
-        public Universe ParseUniverse(string userInput)
+        public static Universe ParseUniverse(string userInput)
         {
             var universe = SplitInput(userInput, ",");
+            var gridWidth = ParseInput(universe[0]);
```
indianbread, line 13: nice, this is much clearer as to what each item in the array corresponds to

ConwaysGameOfLife/InputParser.cs
```diff
@@ -3,32 +3,32 @@
...
-            return new Location(ParseInput(location[0]), ParseInput(location[1]));
+            var x = ParseInput(location[0]);
+            var y = ParseInput(location[1]);
+            ThrowException(x, y);
+            return new Location(x, y);
+        }
+
+        private static void ThrowException(int x, int y)
+        {
+            var message = String.Format("String should only contain numbers greater than or equal to 0.");
```
indianbread, line 30: you don't need to use string.format here

ConwaysGameOfLife/InputParser.cs
```diff
@@ -3,32 +3,32 @@
...
-            return liveCellLocations;
-        }
 
-        public Universe ParseUniverse(string userInput)
+        public static Universe ParseUniverse(string userInput)
         {
             var universe = SplitInput(userInput, ",");
+            var gridWidth = ParseInput(universe[0]);
+            var gridLength = ParseInput(universe[1]);
+            ThrowException(gridWidth, gridLength);
```
indianbread, line 15: rename/refactor this bit so it's clearer that it's making sure that x and y are > 0 before throwing an exception. Alternatively it can go into the validation class as it's validating the width and length - but it depends on what you consider as validating vs parsing - so up to you. Would like your thoughts on this too @Kedoodle

ConwaysGameOfLife/InputParser.cs
```diff
@@ -3,32 +3,32 @@
...
         }
 
-        public Location ParseLocation(string input)
+        public static Location ParseLocation(string input)
         {
             var location = SplitInput(input, ",");
-            return new Location(ParseInput(location[0]), ParseInput(location[1]));
+            var x = ParseInput(location[0]);
+            var y = ParseInput(location[1]);
+            ThrowException(x, y);
```
indianbread, line 24: same as before

indianbread (approved):
nice refactors

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -14,10 +14,9 @@ public GeneratorShould()
         public void WhenInitialized_DisplayUniverse()
```
kedoodle, line 14: Are we doing the action when the Generator is initialised? i.e. is `DisplayUniverse` called in the constructor? Some of the first tests you wrote appropriately had `WhenInitialised` because they were testing some "state" of the object immediately upon initialisation. In this test, we're calling `DisplayUniverse` as part of the test, which is a sign that we're not testing the `Generator`'s initialised state.

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -14,10 +14,9 @@ public GeneratorShould()
...
         {
             var output = new StubOutput();
-            var formatter = new OutputFormatter();
-            var expected = ". . . \n. . . \n. . . \n";
+            var expected = "💀💀💀\n💀💀💀\n💀💀💀\n";
 
-            _generator.DisplayUniverse(output, formatter);
+            _generator.DisplayUniverse(output);
 
             Assert.Equal(expected, output.GetWriteLine());
```
kedoodle, line 21: Are we testing anything here? At what point does the "actual" value get populated? Would this test still pass without line 19?

ConwaysGameOfLife.Tests/InputParserShould.cs
```diff
@@ -6,69 +6,45 @@ namespace ConwaysGameOfLife.Tests
...
-                new Location(0,0),
-                new Location(0,1),
-                new Location(0,2)
-            };
-
-            var result = _parser.ParseLocations(input);
+            var input = "a,4";
 
-            Assert.True(UniverseHelper.ListsOfLocationsAreEqual(expected, result));
+            Assert.Throws<ArgumentException>(() => InputParser.ParseLocation(input));
```
kedoodle, line 27: Is there a more graceful way that we can handle when an invalid string is passed in? Or will this exception be handled elsewhere?

ConwaysGameOfLife.Tests/InputParserShould.cs
kedoodle: Can we have a few more test cases here e.g. what if there is no comma?

ConwaysGameOfLife.Tests/InputParserShould.cs
```diff
@@ -6,69 +6,45 @@ namespace ConwaysGameOfLife.Tests
...
-        [InlineData("3,p", 3, -1)]
-        [InlineData("*,2", -1, 2)]
         public void ReturnLocation_GivenAStringOfOneLocation(string input, int x, int y) 
         {
             var expected = new Location(x, y);
 
-            var result = _parser.ParseLocation(input);
+            var result = InputParser.ParseLocation(input);
 
             Assert.True(UniverseHelper.LocationsAreEqual(expected, result));
```
kedoodle, line 19: Thinking about what you'd see if the test were to fail in different situations e.g. if result weren't a `Location` at all, you may think that `ParseLocation` just wasn't getting the coordinates correctly, but in actuality, `result` is a null or even an entirely different object. Could be worth having an extra, initial assertion on the object type of `result` being `Location`. This would also make the test clearer and truer to the name (which is a good name, by the way).

ConwaysGameOfLife.Tests/OutputFormatterShould.cs
```diff
@@ -5,36 +5,31 @@ namespace ConwaysGameOfLife.Tests
...
-        public OutputFormatterShould()
-        {
-            _formatter = new OutputFormatter();
-        }
         [Theory]
-        [InlineData(3, 3, ". . . \n. . . \n. . . \n")]
-        [InlineData(4, 3, ". . . . \n. . . . \n. . . . \n")]
+        [InlineData(3, 3, "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        [InlineData(4, 3, "💀💀💀💀\n💀💀💀💀\n💀💀💀💀\n")]
         public void WhenGivenDimensions_FormatUniverseAllDeadCells(int gridWidth, int gridLength, string expected)
```
kedoodle, line 11: Can we rejig this name a bit? We don't give the `OutputFormatter` dimensions as parameters to a function, at least not directly. Perhaps relate the name to what kind of `Universe` we're passing to the `FormatUniverse` function in this test.

ConwaysGameOfLife.Tests/OutputFormatterShould.cs
```diff
@@ -5,36 +5,31 @@ namespace ConwaysGameOfLife.Tests
...
-        [InlineData("XXX", "XXX", "XXX", ". . . \n. . . \n. . . \n")]
-        [InlineData("XXX", "OXX", "XXX", ". . . \n* . . \n. . . \n")]
-        [InlineData("OXX", "OXX", "OXX", "* . . \n* . . \n* . . \n")]
-        [InlineData("XXXO", "XXXX", "XXXX", ". . . * \n. . . . \n. . . . \n")]
-        public void WhenGivenDimensionsFromUser_FormatUniverse(string row1, string row2, string row3, string expected)
+        [InlineData("XXX", "XXX", "XXX", "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        [InlineData("XXX", "OXX", "XXX", "💀💀💀\n💟💀💀\n💀💀💀\n")]
+        [InlineData("OXX", "OXX", "OXX", "💟💀💀\n💟💀💀\n💟💀💀\n")]
+        [InlineData("XXXO", "XXXX", "XXXX", "💀💀💀💟\n💀💀💀💀\n💀💀💀💀\n")]
+        public void WhenGivenDimensions_FormatUniverse(string row1, string row2, string row3, string expected)
```
kedoodle, line 25: Same goes with this name - we don't give the `OutputFormatter` dimensions directly. Apologies that this wasn't picked up in previous review when commenting that the dimensions aren't necessarily from a user.

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -44,7 +44,7 @@ public void WhenInitialized_FindCellStateOfCell(CellState state, int x, int y)
         [InlineData("XXX", "OXX", "XXX", 1, 1, 1 )]
         [InlineData("OOO", "OOO", "OOO", 1, 1, 8)]
         [InlineData("OXX", "XXX", "XXO", 1, 1, 2)]
-        public void ReturnCountOfLiveNeighbours_GivenCoordinates(string row1, string row2, string row3, int x, int y, int expected) 
+        public void ReturnCountOfLiveNeighbours_GivenNonBoundaryCoordinates(string row1, string row2, string row3, int x, int y, int expected)
```
kedoodle, line 47: Nice. How about some `Universe`s that are different dimensions? We could theoretically have made our functions only work for 3x3 `Universe` and we wouldn't know.

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -57,6 +57,24 @@ public void ReturnCountOfLiveNeighbours_GivenCoordinates(string row1, string row
             Assert.Equal(expected, result);
         }
 
+        [Theory]
+        [InlineData("XXX", "XXX", "XXX", 0, 1, 0)]
+        [InlineData("XXX", "OXX", "XXX", 1, 2, 1)]
+        [InlineData("OOO", "OOO", "OOO", 2, 1, 8)]
+        [InlineData("OXX", "XXX", "XXO", 2, 1, 2)]
```
kedoodle, line 64: Same here RE: 3x3 only `Universe`

ConwaysGameOfLife.Tests/ValidatorShould.cs
```diff
@@ -20,25 +20,10 @@ public void WhenValidParsedInput_IsUniverse()
...
-        [InlineData(-1, -1)]
-        [InlineData(3, -1)]
-        public void WhenInvalidParsedInput_IsNotUniverse(int x, int y)
+        [Fact]
+        public void NotAcceptUniverse_WhenInputIsInvalid()
         {
-            var input = new Universe(x, y);
+            var input = new Universe(0, 3);
 
             var result = _validator.IsUniverse(input);
```
kedoodle, line 28: Will the `Validator` ever get passed a `Universe` that is invalid given that the `InputParserShould` tests for parsing a `Universe` with invalid input expects an exception to be thrown? These two tests may be competing against each other - the classes they test may be stepping on each others' toes. We need a clear responsibility for each.

ConwaysGameOfLife/Generator.cs
```diff
@@ -38,9 +38,9 @@ public Generator(Universe universe)
             
         }
 
-        public void DisplayUniverse(IOutput output, OutputFormatter formatter)
+        public void DisplayUniverse(IOutput output)
```
kedoodle, line 41: Will the `Generator` be responsible for telling the `IOutput` to do other things as well e.g. output the prompts for user input? Have a think about what "components" we need to build a `Generator` and how they should exist in relation to the `Generator` i.e. what dependency injection will be used.

ConwaysGameOfLife/InputParser.cs
kedoodle: Have put thoughts in a previous comment in the validator tests. Would like clearer separation of responsibilities between the `Validator` and the `InputParser`. I see the `Validator` working as a pre-check to the `InputParser` - attempting to validate after a "successful" parse can only catch when the format is correct (and an object is actually initialised out of a string), but what about when the format is wrong? At the moment, an exception is thrown when the format is wrong - no object is initialised for the `Validator` to validate. In this case, the `InputParser` implicitly does the validation because it throws an exception when the format is wrong, not giving the `Validator` a chance to do any validation.

ConwaysGameOfLife/InputParser.cs
```diff
@@ -44,7 +44,5 @@ private static int ParseInput(string numberInput)
             }
             return -1;
```
kedoodle, line 45: Do we need this wrapper for `Int32.TryParse`? We lose a bit of functionality here, given that if we pass in "-1" vs an invalid `int`, we can't tell the outputs apart.

ConwaysGameOfLife/OutputFormatter.cs
```diff
@@ -1,32 +1,32 @@
...
     {
-        // displays universe & instructions
-        // TODO: try StringBuilder 
+        // formats universe display 
 
-        public string FormatUniverse(Universe universe) // universe created with user input - dimensions & live cells ?? - needed??
+        public static string FormatUniverse(Universe universe) // universe created with user input - dimensions & live cells ?? - needed??
         {
-            var format = "";
+            StringBuilder format = new StringBuilder();
```
kedoodle, line 12: Does the `format` name have any significance? Perhaps we could call it `sb` or `stringBuilder` or `universeBuilder` (examples only - make/pick one that makes sense to you). Also, we can still use the implicit `var` type instead of explicit `StringBuilder`.

```suggestion
            var sb = new StringBuilder();
```

kedoodle (approved):
Great stuff, sorry I didn't do this before the weekend. The `Validator` and `InputParser` are both coming along nicely - I've added some thoughts and suggestions for tweaking their mechanics. Nearly there now with this kata - the `Generator` has a lot of potential now that you have laid out most of the groundwork with the various component classes. It'll all come together very quickly now.

kedoodle:
Also wanted to call out the awesome job with the "boy scout rule". +80 lines and -106 lines this PR along with a load of progress!!

PR #12, opened 2021-06-15
21 06 15

lindajohnstone:


ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -10,8 +10,9 @@ public GeneratorShould()
         {
             _generator = new Generator(new Universe(3, 3));
         }
+
         [Fact]
-        public void WhenInitialized_DisplayUniverse()
+        public void WhenUniverseCreatedInitially_DisplayUniverse() // TODO: if parameter on line 19 is removed, test fails
```
kedoodle, line 15: +1 thanks for clarifying

probably need to inject the `StubOutput` as part of the setup for this test once `DisplayUniverse` no longer takes the `IOutput` as a param

ConwaysGameOfLife.Tests/OutputFormatterShould.cs
```diff
@@ -8,7 +8,7 @@ public class OutputFormatterShould
         [Theory]
         [InlineData(3, 3, "💀💀💀\n💀💀💀\n💀💀💀\n")]
         [InlineData(4, 3, "💀💀💀💀\n💀💀💀💀\n💀💀💀💀\n")]
-        public void WhenGivenDimensions_FormatUniverseAllDeadCells(int gridWidth, int gridLength, string expected)
+        public void WhenUniverseCreatedInitially_FormatUniverseAllDeadCells(int gridWidth, int gridLength, string expected)
```
kedoodle, line 11: Perhaps 
```suggestion
        public void FormatUniverse_GivenUniverseOfAllDeadCells(int gridWidth, int gridLength, string expected)
```

ConwaysGameOfLife.Tests/OutputFormatterShould.cs
```diff
@@ -22,7 +22,7 @@ public void WhenGivenDimensions_FormatUniverseAllDeadCells(int gridWidth, int gr
         [InlineData("XXX", "OXX", "XXX", "💀💀💀\n💟💀💀\n💀💀💀\n")]
         [InlineData("OXX", "OXX", "OXX", "💟💀💀\n💟💀💀\n💟💀💀\n")]
         [InlineData("XXXO", "XXXX", "XXXX", "💀💀💀💟\n💀💀💀💀\n💀💀💀💀\n")]
-        public void WhenGivenDimensions_FormatUniverse(string row1, string row2, string row3, string expected)
+        public void WhenUniverseRecreated_FormatUniverse(string row1, string row2, string row3, string expected)
```
kedoodle, line 25:
```suggestion
        public void FormatUniverse_GivenUniverseOfDeadAndAliveCells_(string row1, string row2, string row3, string expected)
```

ConwaysGameOfLife.Tests/ValidatorShould.cs
```diff
@@ -10,47 +9,55 @@ public ValidatorShould()
         {
             _validator = new Validator();
         }
+
         [Fact]
         public void AcceptUniverse_WhenInputIsValid()
         {
-            var input = new Universe(3, 3);
+            var input = "3,3";
```
kedoodle, line 16: Good opportunity to add some more test cases here by making it a `[Theory]`

ConwaysGameOfLife.Tests/ValidatorShould.cs
```diff
@@ -10,47 +9,55 @@ public ValidatorShould()
         {
             _validator = new Validator();
         }
+
         [Fact]
         public void AcceptUniverse_WhenInputIsValid()
```
kedoodle, line 14: Is this name clearer? (It might not be - ignore if it isn't) If so, could adopt similar names for other tests in `ValidatorShould`
```suggestion
        public void ReturnTrue_GivenValidUniverseInput()
```

ConwaysGameOfLife/Generator.cs
```diff
@@ -29,19 +30,32 @@ rules checked
...
         Universe _universe;
+        IOutput _output;
 
-        
 
         public Generator(Universe universe)
         {
             _universe = universe;
-            
+            _output = new ConsoleOutput();
```
kedoodle, line 39: This will probably need to be passed in as a parameter to `Generator` so you can inject your `StubOutput` when testing

ConwaysGameOfLife/Messages.cs
```diff
@@ -10,6 +10,6 @@ public class Messages
 
         public const string QuitGame = "Enter 'q' to quit Game.";
 
-        public const string RequestLiveCells = "Please enter the x, y coordinates for Live cells."; // TODO: format of string?
+        public const string RequestLiveCells = "Please enter the x, y coordinates for one live cell."; // TODO: format of string?
```
kedoodle, line 13: could be worth giving the user an example? 
```suggestion
        public const string RequestLiveCells = "Please enter the coordinates for one live cell e.g. 0,1."; // TODO: format of string?

```

ConwaysGameOfLife/Validator.cs
```diff
@@ -5,14 +5,45 @@ namespace ConwaysGameOfLife
...
+        
+        public bool IsValidUniverse(string input)
+        {
+            var isNumber = SplitInput(input, ",");
+            if (!input.Contains(",")) return false;
+            if (input.Length > 3) return false;
+            if (isNumber.Length != 2) return false;
+            if (!ParseInput(isNumber[0]) || !ParseInput(isNumber[1])) return false;
+            if (Int32.Parse(isNumber[0]) <= 0) return false;
+            return true;
```
kedoodle, line 26: Probably worth thinking about the ordering of these e.g. the check for contains "," probably could go first. It could also be a count of the `char` `',' == 1`

The `isNumber` variable name could be something like `inputComponents` or something to do with the input but split.

ConwaysGameOfLife/Validator.cs
```diff
@@ -5,14 +5,45 @@ namespace ConwaysGameOfLife
     public class Validator
     {
         // validates input
-        public bool IsUniverse(Universe input)
+
+        public static bool ParseInput(string numberInput)
+        {
+            if (Int32.TryParse(numberInput, out var number))
```
kedoodle, line 11: can use a discard here to make it clear that we're not using the `out var` at all
```suggestion
            if (Int32.TryParse(numberInput, out var _))
```

ConwaysGameOfLife/Validator.cs
kedoodle: is `input.Length` a determining factor for validity? conceivably the universe could be 10x5, for example, in which case `"10,3"` should be valid. could add this as a test case

ConwaysGameOfLife/Validator.cs
```diff
@@ -5,14 +5,45 @@ namespace ConwaysGameOfLife
...
+            if (!input.Contains(",")) return false;
+            if (input.Length > 3) return false;
+            if (isNumber.Length != 2) return false;
+            if (!ParseInput(isNumber[0])) return false;
+            if (!ParseInput(isNumber[1])) return false;
+            if (Int32.Parse(isNumber[0]) < 0) return false;
+            if (Int32.Parse(isNumber[1]) < 0) return false;
+            if (Int32.Parse(isNumber[0]) >= gridWidth) return false;
+            if (Int32.Parse(isNumber[1]) >= gridLength) return false;
+            return true;
```
kedoodle, line 41: same here with the ordering and `char ','`.

As for combining these into a nice boolean, we could use some intermediate variables e.g. (not 100% on syntax as just using the github editor)
```
            var containsComma = input.Contains(",");

            var components = SplitInput(input, ",");
            var hasTwoComponents = components.Length == 2;
            var xIsValid = Int32.TryParse(components[0], out var x) && x >= 0 && x < gridWidth;
            var yIsValid = Int32.TryParse(components[1], out var y) && y >= 0 && y < gridLength;

            return containsComma && hasTwoComponents && xIsValid && yIsValid;
```

kedoodle (approved):
looking good. liking the clarity on what the validator and inputparser each do

ConwaysGameOfLife.Tests/ValidatorShould.cs
```diff
@@ -10,47 +9,55 @@ public ValidatorShould()
...
+        [InlineData("3,3,")]
+        [InlineData("3,3,3")]
+        [InlineData("a,b")]
+        [InlineData("-1,4")]
+        public void NotAcceptUniverse_WhenInputIsInvalid(string input)
         {
-            var input = new Universe(0, 3);
-
-            var result = _validator.IsUniverse(input);
+            var result = _validator.IsValidUniverse(input);
```
indianbread, line 33: nice naming

ConwaysGameOfLife/Generator.cs
indianbread: This is called dependency injection - some [fun reading ](https://alexalvess.medium.com/dependency-injection-and-inversion-of-control-on-net-core-3136fe98b72) to give you a bit more info

ConwaysGameOfLife/Validator.cs
indianbread: TryParse returns true if it can parse successfully, and false if it cannot https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=net-5.0
Alternative options:
1. assign the Int.TryParse result to a variable and return the variable,
2. return it directly
3. call tryparse where you want to use it and don't have the ParseInput method at all

ConwaysGameOfLife/Validator.cs
```diff
@@ -5,14 +5,45 @@ namespace ConwaysGameOfLife
...
+            {
+                return true;
+            }
+            return false;
+        }
+        
+        public bool IsValidUniverse(string input)
+        {
+            var isNumber = SplitInput(input, ",");
+            if (!input.Contains(",")) return false;
```
indianbread, line 21: this will still return true if there is more than 1 comma

ConwaysGameOfLife/Validator.cs
```diff
@@ -5,14 +5,45 @@ namespace ConwaysGameOfLife
...
+            if (input.Length > 3) return false;
+            if (isNumber.Length != 2) return false;
+            if (!ParseInput(isNumber[0]) || !ParseInput(isNumber[1])) return false;
+            if (Int32.Parse(isNumber[0]) <= 0) return false;
```
indianbread, line 25: what about the second number?

indianbread (approved):

PR #14, opened 2021-06-17
21 06 17

lindajohnstone:


ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -22,5 +24,88 @@ public void FormatUniverse_GivenUniverseOfAllDeadCells()
 
             Assert.Equal(expected, _output.GetWriteLine());
         }
+
+        [Theory]
+        [InlineData("3,3", 3, 3)]
+        [InlineData("10,10", 10, 10)]
+        public void SetUniverseDimensions_GivenValidInput(string input, int gridWidth, int gridLength)
```
indianbread, line 31: It's creating a universe here, and not just setting the dimensions

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -22,5 +24,88 @@ public void FormatUniverse_GivenUniverseOfAllDeadCells()
...
+
+            Assert.True(UniverseHelper.UniversesAreEqual(new Universe(gridWidth, gridLength), result));
+            Assert.Equal(gridWidth, result.GridWidth);
+            Assert.Equal(gridLength, result.GridLength);
+        }
+
+        [Theory]
+        [InlineData("3,3", 3, 3)]
+        [InlineData("5,5", 5, 5)] // TODO: this is failing 
+        public void SetUniverseDimensionsInGenerateUniverse_GivenValidInput(string input, int gridWidth, int gridLength)
```
indianbread, line 44: The above test is testing that the universe created by the generator is equal to the expected universe. Part of that check already includes checking the dimensions, so this test may not be required.

ConwaysGameOfLife/Generator.cs
```diff
@@ -55,5 +69,20 @@ public void Run()
 
             }
         }
+
+        public Universe SetUniverseDimensions(string input)
+        {
+            var isValidUniverse = Validator.IsValidUniverse(input);
+            if (isValidUniverse) return InputParser.ParseUniverse(input);
+            throw new InvalidFormatException(String.Format("Invalid input. Please try again." + Environment.NewLine + Messages.RequestDimensions));
```
kedoodle, line 77: A few comments on this:
1. have you tried [string interpolation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) before? it's some nice syntactic sugar that is common in many languages, including C#
2. Is there a nicer way to handle how the program will re-request user dimensions? Not sure about having this exception message
3. Following on from point 2., is this function following command-query separation? The name implies that we are... but then the function actually doesn't set anything.

ConwaysGameOfLife/Generator.cs
```diff
@@ -55,5 +69,20 @@ public void Run()
...
         }
+
+        public Universe SetUniverseDimensions(string input)
+        {
+            var isValidUniverse = Validator.IsValidUniverse(input);
+            if (isValidUniverse) return InputParser.ParseUniverse(input);
+            throw new InvalidFormatException(String.Format("Invalid input. Please try again." + Environment.NewLine + Messages.RequestDimensions));
+        }
+
+        public Location SetLiveCellLocation(string input, int gridWidth, int gridLength)
```
kedoodle, line 80: read into CQS, then revisit this function name

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -22,5 +24,88 @@ public void FormatUniverse_GivenUniverseOfAllDeadCells()
...
+        [InlineData("a,b")]
+        [InlineData("-1,4")]
+        [InlineData("10,0")]
+        public void ThrowException_GivenInvalidUniverseDimensions(string input)
+        {
+            Assert.Throws<InvalidFormatException>(() => _generator.SetUniverseDimensions(input)); ;
+        }
+
+        [Fact]
+        public void FormatUniverse_FromValidInput()
```
indianbread, line 69: No need to test this here - you are basically testing that the output is formatted correctly which has already been done in your output formatter tests

ConwaysGameOfLife.Tests/StubInput.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            _queue = new Queue<string>();
+        }      
+
+        public string ReadLine()
+        {
+            return _queue.Dequeue();
+        }
+        
+        public void WithReadLine(string value)
```
kedoodle, line 19: nice use of queues. can we think of a better name for this function? it's purpose is to set up the `StubInput` with some input that will be read in later when the `IInput.ReadLine` function is called.

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -22,5 +24,88 @@ public void FormatUniverse_GivenUniverseOfAllDeadCells()
...
+        [InlineData("0,0", 0, 0, 3, 3)]
+        [InlineData("0,0", 0, 0, 4, 3)]
+        [InlineData("2,1", 2, 1, 10, 5)]
+        [InlineData("2,1", 2, 1, 4, 6)]
+        public void SetAliveCellLocation_GivenValidInput(string input, int x, int y, int gridWidth, int gridLength)
+        {
+            var universe = new Universe(gridWidth, gridLength);
+            var result = _generator.SetLiveCellLocation(input, universe.GridWidth, universe.GridLength);
+
+            Assert.True(UniverseHelper.LocationsAreEqual(new Location(x, y), result));
```
indianbread, line 90: What about the cell state?

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -22,5 +24,88 @@ public void FormatUniverse_GivenUniverseOfAllDeadCells()
...
             Assert.Equal(expected, _output.GetWriteLine());
         }
+
+        [Theory]
+        [InlineData("3,3", 3, 3)]
+        [InlineData("10,10", 10, 10)]
+        public void SetUniverseDimensions_GivenValidInput(string input, int gridWidth, int gridLength)
+        {
+            _input.WithReadLine(input);
+            var result = _generator.SetUniverseDimensions(input);
```
kedoodle, line 34: see notes on CQS in `Generator` comments. are we actually doing a `Set` here? no change in the `_generator` from what I can tell

ConwaysGameOfLife.Tests/ValidatorShould.cs
```diff
@@ -4,18 +4,12 @@ namespace ConwaysGameOfLife.Tests
...
-        public ValidatorShould()
-        {
-            _validator = new Validator();
-        }
-
         [Theory]
         [InlineData("3,3")]
         [InlineData("10,3")]
-        public void ReturnTrue_GivenValidUniverseInputputIsValid(string input)
+        public void ReturnTrue_GivenValidUniverseInputput(string input)
```
indianbread, line 10:
```suggestion
        public void ReturnTrue_GivenValidUniverseInput(string input)

```

ConwaysGameOfLife/Generator.cs
```diff
@@ -27,19 +27,33 @@ rules checked
...
             _output = output;
+            _input = input;
         }
 
+        public void GenerateUniverse()
+        // if lines uncommented, both tests (line 44) fail. is this method doing too much? how to test? - does it need to be tested
+        {
+            _output.WriteLine(Messages.RequestDimensions);
+            var universeInput = _input.ReadLine();
+            var universe = SetUniverseDimensions(universeInput);
```
kedoodle, line 47: Is this doing what we think it is?

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -22,5 +24,88 @@ public void FormatUniverse_GivenUniverseOfAllDeadCells()
...
+            Assert.Equal(gridLength, result.GridLength);
+        }
+
+        [Theory]
+        [InlineData("3,3", 3, 3)]
+        [InlineData("5,5", 5, 5)] // TODO: this is failing 
+        public void SetUniverseDimensionsInGenerateUniverse_GivenValidInput(string input, int gridWidth, int gridLength)
+        {
+            _input.WithReadLine(input);
+            _generator.GenerateUniverse();
```
kedoodle, line 47: again, are we storing the universe that we're generating?

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -22,5 +24,88 @@ public void FormatUniverse_GivenUniverseOfAllDeadCells()
...
+            _input.WithReadLine(input);
+            var result = _generator.SetUniverseDimensions(input);
+
+            Assert.True(UniverseHelper.UniversesAreEqual(new Universe(gridWidth, gridLength), result));
+            Assert.Equal(gridWidth, result.GridWidth);
+            Assert.Equal(gridLength, result.GridLength);
+        }
+
+        [Theory]
+        [InlineData("3,3", 3, 3)]
```
kedoodle, line 42: this is passing because of the test constructor passing in `new Universe(3, 3)`. be careful with using "similar" test cases frequently i.e. mostly using 3x3 universes. it can lead to tests passing when they shouldn't. luckily, this test is a `[Theory]` so you can tell that the overall function isn't working as intended even though the 3x3 test passed

ConwaysGameOfLife/Generator.cs
```diff
@@ -27,19 +27,33 @@ rules checked
...
+        public Generator(Universe universe, IOutput output, IInput input)
         {
-            _universe = universe;
+            Universe = universe;
             _output = output;
+            _input = input;
         }
 
+        public void GenerateUniverse()
+        // if lines uncommented, both tests (line 44) fail. is this method doing too much? how to test? - does it need to be tested
```
indianbread, line 43: one approach is to think of this in 2 stages - 1st stage is creating the universe with all dead cells, second stage is setting up the starter universe with live cells. lines 48 onwards can go in a separate function as it's for the second stage.

kedoodle (approved):
nice. coming together nicely. especially like the `StubInput` queue implementation of `IInput`

overall thoughts:
- is the generator doing what it's supposed to? was the initial intention for it to take a `Universe`, apply various `Rules`, and return a new `Universe`? 
- perhaps the generator could be renamed to better reflect it's current purpose, and a new `Generator` class could be made which actually generates the next iteration of a `Universe`

indianbread (approved):

PR #15, opened 2021-06-18
21/ 06 18

lindajohnstone:


ConwaysGameOfLife/GameController.cs
```diff
@@ -0,0 +1,112 @@
...
+            var input = _input.ReadLine();
+            var isValidUniverse = Validator.IsValidUniverse(input);
+            while (!isValidUniverse)
+            {
+                InvalidUniverseString(input);
+            }
+            return input;
+        }
+
+        private void InvalidUniverseString(string input)
```
indianbread, line 89: I think it's better if you just put the code within this function directly in the while loop as it's doing 2 different things

ConwaysGameOfLife/GameController.cs
```diff
@@ -0,0 +1,112 @@
...
+            var input = _input.ReadLine();
+            var isValidLocation = Validator.IsValidLocation(input, _universe.GridWidth, _universe.GridLength);
+            while (!isValidLocation) 
+            {
+                InvalidLocationString(input);
+            }
+            return input;
+        }
+
+        private void InvalidLocationString(string input)
```
indianbread, line 106: same here

indianbread (approved):

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -0,0 +1,72 @@
...
+            var controller = new GameController(universe, _output, _input);
+            _input.GetReadLine(input);
+            var expected = input;
+
+            var result = controller.CreateValidUniverseString();
+
+            Assert.Equal(expected, result);
+        }
+
+        //TODO: how to test error message in CreateValidUniverseString
```
kedoodle, line 44: Probably want to have a separate test for this and preload the `StubInput` with different inputs so that when your `CreateValidUniverseString` function is called in each test, the inputs are whatever you want them to be.

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        StubInput _input;
+
+        public GameControllerShould()
+        {
+            _output = new StubOutput();
+            _input = new StubInput();
+        }
+        [Fact]
+        public void FormatUniverse_GivenUniverseOfAllDeadCells()
```
kedoodle, line 18: Are we testing the behaviour of the `GameController` or are we testing dependencies? e.g. we already have tests for `FormatUniverse`. The test itself is probably okay (if we are okay with having `DisplayUniverse` being a public function), just the name could be clearer that we're expecting a `Universe` being written to the `IOutput`

ConwaysGameOfLife/GameController.cs
```diff
@@ -0,0 +1,112 @@
...
+            universe
+            rules
+        */
+        private Universe _universe;
+
+        private IOutput _output;
+
+        private IInput _input;
+
+        public GameController(Universe universe, IOutput output, IInput input)
```
kedoodle, line 36: Are you sure you want a `Universe` being passed into the `GameController`? Would the `GameController` not make the very first `Universe` from the `IInput`?

A nitpick, but input/output generally is shortened to I/O - could consider having `input` as the first parameter and `output` as the second.
```suggestion
        public GameController(IInput input, IOutput output)
```

ConwaysGameOfLife/GameController.cs
```diff
@@ -0,0 +1,112 @@
...
+            _output = output;
+            _input = input;
+        }
+
+        public void CreateInitialUniverse()
+        {
+            _output.WriteLine(Messages.RequestDimensions);
+            var input = CreateValidUniverseString();
+            InputParser.ParseUniverse(input);
+            DisplayUniverse();
```
kedoodle, line 48: Displaying the created `Universe` may be considered a "side-effect" i.e. the purpose of the function would be fulfilled without actually displaying the `Universe`. Consider explicitly calling `DisplayUniverse` when needed in the `Run` function.
```suggestion
```

ConwaysGameOfLife/GameController.cs
kedoodle: Generally wouldn't comment for resource optimisation, but in this case, you'd be unnecessarily creating recursive calls to `CreateValidUniverseString` whenever the input is invalid. This could, however unlikely, lead to a memory problem as you're storing the `input` and `isValidUniverse` values for each recursive call until you finally get valid input. Here, another problem is presented because we're not only making recursive calls, but we're also using a while loop. The while loop will loop forever because we never actually save anything inside and reevaluate the predicate!

For demonstration:
1. `CreateValidUniverseString` is called
2. `input` is assigned some invalid input `"invalid1"`
3. `isValidUniverse` is evaluated to false
4. `!isValidUniverse` is evaluiated to true, we go into the while loop
5. `InvalidUniverseString("invalid1")` is called
6. We write to the output to request dimensions again
7. `CreateValidUniverseString` is called again, repeating steps 1-6 but potentially with any invalid input; until
8. `CreateValidUniverseString` is called for the final time
9. `input`  is assigned some valid input `"3,3"`
10. `isValidUniverse` is evaluated to true
11. `!isValidUniverse` is evaluated to false
12. `input` is returned to the latest `CreateValidUniverseString` call from 7.
13. We're now back at the end of the latest `CreateValidUniverseString` call's while loop
14. `!isValidUniverse` gets evaluated to true (!!!) because the `input` from 12. isn't stored and reevaluated
15. `InvalidUniverseString` is called again 
16. We're just stuck in a cycle even if the user repeatedly puts valid input

Recommendations:
- in general, choose only one of recursion or while loop for any given function
  - for this particular function, recursion gives no benefit (we can assign some reading if you're interested in recursion and where it might be useful)
- as @indianbread suggested, move the `InvalidUniverseString` lines 91 and 92 directly inside the while loop of the `CreateValidUniversString`. This should help clarify the demonstration above.
- once you understand the demonstration:
  1. change the code from line 92 to just read a new line in like on line 80
  2. add another line to reevaluate `isValidUniverse` with the new input
  3. let the while loop do the hard work until some valid input is reached! 🎉

ConwaysGameOfLife/GameController.cs
```diff
@@ -0,0 +1,112 @@
...
+            while (!isValidUniverse)
+            {
+                InvalidUniverseString(input);
+            }
+            return input;
+        }
+
+        private void InvalidUniverseString(string input)
+        {
+            _output.WriteLine($"Invalid input. Please try again.{Environment.NewLine}{Messages.RequestDimensions}");
```
kedoodle, line 91: these `_output.WriteLine` calls are cheap. feel free to use them more frequently if it aids readability.

```suggestion
            _output.WriteLine("Invalid input. Please try again.");
            _output.WriteLine(Messages.RequestDimensions);
```

ConwaysGameOfLife/GameController.cs
kedoodle: see the long comment from above. same thing applies to this and `CreateValidLocationString`

ConwaysGameOfLife/GameController.cs
```diff
@@ -0,0 +1,112 @@
...
+            while (!isValidLocation) 
+            {
+                InvalidLocationString(input);
+            }
+            return input;
+        }
+
+        private void InvalidLocationString(string input)
+        {
+            _output.WriteLine($"Invalid input. Please try again.{Environment.NewLine}{Messages.RequestLiveCell}");
```
kedoodle, line 108:
```suggestion
            _output.WriteLine(Messages.InvalidInput);
            _output.WriteLine(Messages.RequestLiveCell);
```

`"Invalid input. Please try again."` is used in your `InvalidUniverseString` function. Can create another `Messages` `const` (I've given it an example name `InvalidInput`

ConwaysGameOfLife/Messages.cs
```diff
@@ -10,8 +10,8 @@ public class Messages
 
         public const string QuitGame = "Enter 'q' to quit Game.";
 
-        public const string Run = "Enter 'p' to play game.";
+        public const string Play = "enter 'p' to play game.";
```
kedoodle, line 13: Why lowercase `enter`?

ConwaysGameOfLife/Messages.cs
```diff
@@ -10,8 +10,8 @@ public class Messages
 
         public const string QuitGame = "Enter 'q' to quit Game.";
 
-        public const string Run = "Enter 'p' to play game.";
+        public const string Play = "enter 'p' to play game.";
 
-        public const string RequestLiveCell = "Please enter the x, y coordinates for one live cell (e.g. '0,1')."; 
+        public const string RequestLiveCell = "Please enter the x, y coordinates for one live cell (e.g. '0,1') ";
```
kedoodle, line 15: Here it doesn't matter so much given that it's just for a `const string`, but consistency is a good practice in code. Humans (especially software developers) are good at picking up patterns i.e. consistency. In this instance, there's a period at the end of all of the other `const string` values.

Speculating, but you may have removed it because it's close to the example? Perhaps removing the parentheses instead may make it less crowded?
```suggestion
        public const string RequestLiveCell = "Please enter the x, y coordinates for one live cell e.g. '0,1'."; 
```

ConwaysGameOfLife/Messages.cs
```diff
@@ -10,8 +10,8 @@ public class Messages
 
         public const string QuitGame = "Enter 'q' to quit Game.";
```
kedoodle, line 11: Feel free to add a bit of flair in your messages to make the kata your own e.g.
```suggestion
        public const string QuitGame = "Seen enough? Enter 'q' to end the Game of Life.";
```

As a side note, you can have `"` as part of your string if you want `` `q` `` to be `"q"` instead. This is done using an escape character (`\` in C# e.g. `\n` prints a newline instead of literally `n`). You can also use verbatim literals, but this is less common. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim

kedoodle (approved):
I've added some more in-depth comments, but don't mistake them for negativity; the `GameController` is actually coming together very nicely. The `Generator` is also a nice "unit" of work now, so you can definitely unit test it more easily when you feel like taking a crack at it. 🥇

ConwaysGameOfLife/Messages.cs
lindajohnstone: Logic (in code) for PopulateUniverseWithLiveCells:
            // _output.Write(Messages.RequestLiveCell);
            // _output.WriteLine($"or {Messages.Play});
            // loop
            // var locationInput = _input.ReadLine();
            // var location = SetLiveCellLocation(locationInput, universe.GridWidth, universe.GridLength);
            // universe.SwitchCellState(universe.GetCellAtLocation(location.X, location.Y));
            // DisplayUniverse();
Because it is the second part of a 'sentence'

PR #16, opened 2021-06-23
21 06 23

lindajohnstone:
GameControllerShould line 100 - input throws an ThrowOverflowOrFormatException. Would like suggestions as to how to test this. I know I will need to catch this as well as InvalidOperationException if the universe is not correctly formatted when testing invalid location strings (line 119).

There are a lot of public methods on GameController. If I make the methods private, do I simply delete the tests or is it possible to test through the Run method which will remain public.

### DIFF
--- a/ConwaysGameOfLife.Tests/GameControllerShould.cs
+++ b/ConwaysGameOfLife.Tests/GameControllerShould.cs
@@ -72,13 +72,13 @@ namespace ConwaysGameOfLife.Tests
 
 
         [Theory]
-        [InlineData("0,0", 3, 3)]
-        [InlineData("0,0", 4, 3)]
-        [InlineData("2,1", 10, 5)]
-        [InlineData("2,1", 4, 6)]
-        public void ReturnLocationString_GivenValidInput(string input, int gridWidth, int gridLength)
+        [InlineData("0,0", "3,3", 3, 3)]
+        [InlineData("0,0", "4,3", 4, 3)]
+        [InlineData("2,1", "10,5", 10, 5)]
+        [InlineData("2,1", "4,6", 4, 6)]
+        public void ReturnLocationString_GivenValidInput(string input, string universeInput, int gridWidth, int gridLength)
         {
-            InputParser.ParseUniverse(input);
+            InputParser.ParseUniverse(universeInput);
             _input.GetReadLine(input);
             var expected = input;
 
@@ -87,6 +87,43 @@ namespace ConwaysGameOfLife.Tests
             Assert.Equal(expected, result);
         }
 
-        //TODO: how to test error message in CreateValidLocationString
+        [Theory]
+        [InlineData(new string[] { "0,3" }, "2,2", "3,3", 3, 3)]
+        [InlineData(new string[] { "3,0" }, "0,0", "3,3" , 3, 3)]
+        [InlineData(new string[] { ",," }, "2,2", "3,3", 3, 3)]
+        [InlineData(new string[] { "3,3," }, "3,2", "4,3", 4, 3)]
+        [InlineData(new string[] { "3,3,3" },"3,2", "4,3", 4, 3)]
+        [InlineData(new string[] { "a,b" }, "1,1", "3,3", 3, 3)]
+        [InlineData(new string[] { "-1,4" }, "2,0", "3,3", 3, 3)]
+        [InlineData(new string[] { "4,-1" }, "2,2", "3,3", 3, 3)]
+        //[InlineData(new string[] { "4 4" }, "2,1", "4 3", 4, 3)] // ThrowOverflowOrFormatException
+        public void ReturnLocationString_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput, string universeString, int gridWidth, int gridLength)
+        {
+            InputParser.ParseUniverse(universeString);
+            _input.GetReadLine(invalidInputs);
+            _input.GetReadLine(validInput);
+            var expectedOutputLine1 = "Invalid input. Please try again.";
+            var expectedOutputLine2 = "Please enter the x, y coordinates for one live cell (e.g. '0,1').";
+
+            var result = _controller.CreateValidLocationString(gridWidth, gridLength);
+            var actualOutputLine1 = _output.GetWriteLine(0);
+            var actualOutputLine2 = _output.GetWriteLine(1);
+
+            Assert.Equal(validInput, result);
+            Assert.Equal(expectedOutputLine1, actualOutputLine1);
+            Assert.Equal(expectedOutputLine2, actualOutputLine2);
+        }
+
+        [Theory]
+        [InlineData(new string[] { "0,0" }, "-9,0", "0,0", -10, 3)] 
+        [InlineData(new string[] { "0,0" }, "-9,0", "0,0", 3, -10)] 
+        public void ThrowInvalidOperationException_GivenInvalidLocationInputsFollowedByValidLocationInputWithInvalidUniverseInput(IEnumerable<string> invalidInputs, string validInput, string universeString, int gridWidth, int gridLength)
+        {
+            InputParser.ParseUniverse(universeString);
+            _input.GetReadLine(invalidInputs);
+            _input.GetReadLine(validInput);
+    
+            Assert.Throws<InvalidOperationException>(() => _controller.CreateValidLocationString(gridWidth, gridLength));
+        }
     }
 }
--- a/ConwaysGameOfLife.Tests/GameControllerShould.cs
+++ b/ConwaysGameOfLife.Tests/GameControllerShould.cs
@@ -9,22 +9,24 @@ namespace ConwaysGameOfLife.Tests
 
         StubInput _input;
 
+        GameController _controller;
         public GameControllerShould()
         {
             _output = new StubOutput();
             _input = new StubInput();
+            _controller = new GameController(_input, _output);
         }
         [Fact]
-        public void DisplayUniverse_GivenUniverseOfAllDeadCells() 
+        public void DisplayUniverse_GivenSuccessfulCreationOfInitialUniverse() 
         {
             var input = "3,3";
-            var controller = new GameController(_input, _output);
-            InputParser.ParseUniverse(input);
+            _input.GetReadLine(input);
+            _controller.CreateInitialUniverse();
             var expected = "💀💀💀\n💀💀💀\n💀💀💀\n";
 
-            controller.DisplayUniverse();
-
-            Assert.Equal(expected, _output.GetWriteLine());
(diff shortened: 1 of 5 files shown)


ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
-        [InlineData("3,3", 3, 3)]
-        [InlineData("10,10", 10, 10)]
-        public void ReturnUniverseString_GivenValidInput(string input, int gridWidth, int gridLength)
+        [InlineData("3,3")]
+        [InlineData("10,10")]
+        public void ReturnUniverseString_GivenValidInput(string input)
         {
-            var universe = new Universe(gridWidth, gridLength);
-            var controller = new GameController(universe, _output, _input);
+            InputParser.ParseUniverse(input);
```
kedoodle, line 40: Don't think this line is needed
```suggestion
```

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
+        public void ReturnUniverseString_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput)
+        {
+            _input.GetReadLine(invalidInputs);
+            _input.GetReadLine(validInput);
+            var expectedOutputLine1 = "Invalid input. Please try again.";
+            var expectedOutputLine2 = "Please enter the width & length for the Universe as a number followed by a comma then a number (e.g. '0,0').";
+
+            var result = _controller.CreateValidUniverseString();
+            var actualOutputLine1 = _output.GetWriteLine(0);
+            var actualOutputLine2 = _output.GetWriteLine(1);
```
kedoodle, line 67: It would be nice to use dynamic indices for these two so that it's testing against the output from the last invalid input. At the moment it's testing the output only the first invalid input.

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
+        public void ReturnLocationString_GivenValidInput(string input, string universeInput, int gridWidth, int gridLength)
         {
-            var universe = new Universe(gridWidth, gridLength);
-            var controller = new GameController(universe, _output, _input);
+            InputParser.ParseUniverse(universeInput);
             _input.GetReadLine(input);
             var expected = input;
 
-            var result = controller.CreateValidLocationString();
+            var result = _controller.CreateValidLocationString(gridWidth, gridLength);
```
kedoodle, line 86: Can you think of a way to make this work? pt. 2

```suggestion
            var result = _controller.CreateValidLocationString();
```

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
-        public void ReturnLocationString_GivenValidInput(string input, int gridWidth, int gridLength)
+        [InlineData("0,0", "3,3", 3, 3)]
+        [InlineData("0,0", "4,3", 4, 3)]
+        [InlineData("2,1", "10,5", 10, 5)]
+        [InlineData("2,1", "4,6", 4, 6)]
+        public void ReturnLocationString_GivenValidInput(string input, string universeInput, int gridWidth, int gridLength)
         {
-            var universe = new Universe(gridWidth, gridLength);
-            var controller = new GameController(universe, _output, _input);
+            InputParser.ParseUniverse(universeInput);
```
kedoodle, line 82: Again, not doing anything here

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
-        [InlineData("0,0", 3, 3)]
-        [InlineData("0,0", 4, 3)]
-        [InlineData("2,1", 10, 5)]
-        [InlineData("2,1", 4, 6)]
-        public void ReturnLocationString_GivenValidInput(string input, int gridWidth, int gridLength)
+        [InlineData("0,0", "3,3", 3, 3)]
+        [InlineData("0,0", "4,3", 4, 3)]
+        [InlineData("2,1", "10,5", 10, 5)]
+        [InlineData("2,1", "4,6", 4, 6)]
+        public void ReturnLocationString_GivenValidInput(string input, string universeInput, int gridWidth, int gridLength)
```
kedoodle, line 80: Can you think of a way to make this work? pt. 1

```suggestion
        public void ReturnLocationString_GivenValidInput(string input, string universeInput)
```

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
+        [Theory]
+        [InlineData(new string[] { "0,3" }, "2,2", "3,3", 3, 3)]
+        [InlineData(new string[] { "3,0" }, "0,0", "3,3" , 3, 3)]
+        [InlineData(new string[] { ",,", ";;" }, "2,2", "3,3", 3, 3)]
+        [InlineData(new string[] { "3,3,", "3,4;" }, "3,2", "4,3", 4, 3)]
+        [InlineData(new string[] { "3,3,3" },"3,2", "4,3", 4, 3)]
+        [InlineData(new string[] { "a,b" }, "1,1", "3,3", 3, 3)]
+        [InlineData(new string[] { "-1,4" }, "2,0", "3,3", 3, 3)]
+        [InlineData(new string[] { "4,-1" }, "2,2", "3,3", 3, 3)]
+        //[InlineData(new string[] { "4 4" }, "2,1", "4 3", 4, 3)] // ThrowOverflowOrFormatException
```
kedoodle, line 100: Remember that the `CreateValidLocationString` function won't return until a valid input is read in. Any error you're getting with this test is because the `StubInput` has run dequeued all of the inputs that you preloaded.

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
+        public void ReturnLocationString_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput, string universeInput, int gridWidth, int gridLength)
+        {
+            InputParser.ParseUniverse(universeInput);
```
kedoodle, line 103: Doing anything?

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
-        [Fact]
-        public void testName()
+        [Theory]
+        [InlineData(new string[] { "0,0" }, "-9,0", "0,0", -10, 3)] 
+        [InlineData(new string[] { "0,0" }, "-9,0", "0,0", 3, -10)] 
+        public void ThrowInvalidOperationException_GivenInvalidLocationInputsFollowedByValidLocationInputWithInvalidUniverseInput(IEnumerable<string> invalidInputs, string validInput, string universeString, int gridWidth, int gridLength)
         {
-            var universe = new Universe(3, 3);
-            Assert.Equal(3, universe.GridWidth);
+            InputParser.ParseUniverse(universeString);
```
kedoodle, line 123: Doing anything?

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -9,64 +10,121 @@ public class GameControllerShould
...
+            _input.GetReadLine(invalidInputs);
+            _input.GetReadLine(validInput);
+    
+            Assert.Throws<InvalidOperationException>(() => _controller.CreateValidLocationString(gridWidth, gridLength));
```
kedoodle, line 127: not sure what this is trying to test. you should never get to the `CreateValidLocationString` without a valid `Universe` already being parsed and stored in `_universe`

ConwaysGameOfLife.Tests/StubInput.cs
```diff
@@ -20,5 +20,13 @@ public void GetReadLine(string value)
         {
             _queue.Enqueue(value);
         }
+
+        public void GetReadLine(IEnumerable<string> values)
+        {
+            foreach (var value in values)
+            {
+                GetReadLine(value);
+            }
```
kedoodle, line 29: side exercise: try make this a `Linq` function

ConwaysGameOfLife/GameController.cs
```diff
@@ -29,34 +29,31 @@ rules checked
...
 
         public void PopulateUniverseWithLiveCells()
         {
             // _output.Write(Messages.RequestLiveCell);
-            // _output.WriteLine($"Or {Messages.Play});
+            // _output.WriteLine($ or {Messages.Play});
             // loop
             // var locationInput = _input.ReadLine();
-            // var location = SetLiveCellLocation(locationInput, universe.GridWidth, universe.GridLength);
+            // var location = SetLiveCellLocation(locationInput, _universe.GridWidth, _universe.GridLength);
```
kedoodle, line 55: do you still need the width and length params? or can you access them within the function

```suggestion
            // var location = SetLiveCellLocation(locationInput);
```

ConwaysGameOfLife/GameController.cs
```diff
@@ -70,43 +67,40 @@ public void Run()
             _output.WriteLine(Messages.Welcome);
             // create initial universe
             CreateInitialUniverse();
```
kedoodle, line 69: We should prompt the user for input either here or first thing within `CreateInitialUniverse`
```suggestion
            // prompt
            CreateInitialUniverse();
```

kedoodle (approved):
> There are a lot of public methods on GameController. If I make the methods private, do I simply delete the tests or is it possible to test through the Run method which will remain public.

possibly want some end to end tests of the `Run` method rather than unit tests of smaller function. All functions should be implicitly tested in the end to end tests.

ConwaysGameOfLife/GameController.cs
```diff
@@ -70,43 +67,40 @@ public void Run()
...
+            DisplayUniverse();
             // add live cells to universe until user presses 'p' to play
+            // PopulateUniverseWithLiveCells();
+            // DisplayUniverse();
             // generator checks all cells if change of state required & creates next universe
             // loop last step until user presses 'q' to quit or all cells are dead
         }
-        
+
         public string CreateValidUniverseString()
```
indianbread, line 78: this method can be private. When you test the CreateInitialUniverse method, it will also be testing this method as it's used in that method. In saying that, see Keddy's comments about testing the run method. If you decide to make this method private too, you can still test all the functionality by testing the run method.

indianbread (approved):

PR #18, opened 2021-06-25
21 06 25

lindajohnstone:
The last test fails - seems to switch the x & y values. Debugged & that is where the issue seems to be. Can we talk about this on Monday?

Input was "2,0"

Expected: 💀💀💀\n💀💀💀\n💟💀💀\n
Actual:   💀💀💟\n💀💀💀\n💀💀💀\n

### DIFF
--- a/ConwaysGameOfLife.Tests/GameControllerShould.cs
+++ b/ConwaysGameOfLife.Tests/GameControllerShould.cs
@@ -89,16 +89,16 @@ namespace ConwaysGameOfLife.Tests
         //     var universeInput = "3,3";
         //     var input = "1,2";
         //     _input.GetReadLine(universeInput);
-        //     _controller.CreateInitialUniverse();
+        //     _controller.CreateInitialUniverse(universeInput);
         //     _input.GetReadLine(input);
-        //     var locationString = _controller.CreateValidLocationString();
-        //     var location = InputParser.ParseLocation(locationString);
+        //     _controller.CreateValidLocationString(input);
+        //     var location = InputParser.ParseLocation(input);
 
         //     var result = _controller.ReturnUniverseAfterSettingLiveCellLocation(location);
 
         //     Assert.Equal(CellState.Alive, result.GetCellAtLocation(location).CellState);
-        // }
-
+        // }  
+             
         [Fact]
         public void EndRunImmediately_GivenQInput()
         {
@@ -115,6 +115,7 @@ namespace ConwaysGameOfLife.Tests
         [InlineData("3,3", "💀💀💀\n💀💀💀\n💀💀💀\n")]
         [InlineData("3,2", "💀💀💀\n💀💀💀\n")]
         [InlineData("9,9", "💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n")]
+        [InlineData("3,4", "💀💀💀\n💀💀💀\n💀💀💀\n💀💀💀\n")]
         public void DisplayUniverse_GivenValidUniverseString(string input, string expected)
         {
             _input.GetReadLine(input);
@@ -139,6 +140,22 @@ namespace ConwaysGameOfLife.Tests
         {
             _input.GetReadLine(invalidInputs);
             _input.GetReadLine(validInput);
+            _input.GetReadLine("q");
+
+            _controller.Run();
+            var actual = _output.GetLastWriteLine();
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Theory]
+        [InlineData("3,3", "1,2", "💀💀💀\n💀💀💟\n💀💀💀\n")] // failing - displays live cell @ location 1,2
+        [InlineData("3,3", "2,0", "💀💀💀\n💀💀💀\n💟💀💀\n")] // failing - displays live cell @ location 0,2
+        [InlineData("3,2", "0,0", "💟💀💀\n💀💀💀\n")]
+        public void DisplayUniverse_GivenValidLocationString(string universeInput, string locationInput, string expected)
+        {
+            _input.GetReadLine(universeInput);
+            _input.GetReadLine(locationInput);
             _input.GetReadLine("q");
 
             _controller.Run();
--- a/ConwaysGameOfLife.Tests/GameControllerShould.cs
+++ b/ConwaysGameOfLife.Tests/GameControllerShould.cs
@@ -33,38 +33,6 @@ namespace ConwaysGameOfLife.Tests
         //     Assert.Equal(expected, actual);
         // }
 
-        
-
-        // [Theory]
-        // [InlineData(new string[] { "0,3", "3,0" }, "1,1")]
-        // [InlineData(new string[] { "0;3", "3-0", "4 4" }, "3,3")]
-        // [InlineData(new string[] { ",," }, "10,10")]
-        // [InlineData(new string[] { "3,3," }, "10,10")]
-        // [InlineData(new string[] { "3,3,3", "3;3;" }, "3,3")]
-        // [InlineData(new string[] { "a,b" }, "3,3")]
-        // [InlineData(new string[] { "-1,4", "4,-1" }, "3,3")]
-        // [InlineData(new string[] { "10,0", "0,10" }, "3,3")]
-        // public void ReturnUniverseString_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput)
-        // {
-        //     _input.GetReadLine(invalidInputs);
-        //     _input.GetReadLine(validInput);
-        //     var invalidOutputMessageLine1 = "Invalid input. Please try again.";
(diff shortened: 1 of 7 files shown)


ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -18,128 +18,150 @@ public GameControllerShould()
...
-                Assert.Equal(invalidOutputMessageLine2, actualOutputLine2);
-            }
-
-            Assert.Equal(validInput, result);
-        }
-
-        [Fact]
-        public void SetLiveCellLocation()
+        [InlineData("3,3", "1,2", "💀💀💀\n💀💀💟\n💀💀💀\n")] // failing - displays live cell @ location 1,2
+        [InlineData("3,3", "2,0", "💀💀💀\n💀💀💀\n💟💀💀\n")] // failing - displays live cell @ location 0,2
```
kedoodle, line 153:
```suggestion
        [InlineData("3,3", "1,2", "💀💀💀\n💀💀💀\n💀💟💀\n")]
        [InlineData("3,3", "2,0", "💀💀💟\n💀💀💀\n💀💀💀\n")]
```
it's the expected value that seems to be off


x,y so 1,2 would be 
💀💀💀
💀💀💀
💀💟💀

and 2,0 would be
💀💀💟
💀💀💀
💀💀💀

ConwaysGameOfLife.Tests/UniverseShould.cs
```diff
@@ -26,7 +26,7 @@ public void WhenInitialized_FindCellStateOfCell(Cell cell)
             var universe = new Universe(3, 3);
             var expected = CellState.Dead;
 
-            var result = universe.GetCellAtLocation(cell.Location.X, cell.Location.Y).CellState;
+            var result = universe.GetCellAtLocation(cell.Location).CellState;
```
kedoodle, line 29: try not to "reach in" too far i.e. too many `.` in one line

```suggestion
            var cell = universe.GetCellAtLocation(cell.Location);
            var result = cell.CellState;
```

ConwaysGameOfLife/GameController.cs
```diff
@@ -42,19 +42,57 @@ public GameController(IInput input, IOutput output)
...
+            if (UserEndsGame(input)) return;
+
+            var isValidUniverse = Validator.IsValidUniverse(input);
+            while (!isValidUniverse)
+            {
+                _output.WriteLine(Messages.InvalidInput);
+                _output.WriteLine(Messages.RequestDimensions);
+                input = _input.ReadLine();
+                isValidUniverse = Validator.IsValidUniverse(input);
+            }
```
kedoodle, line 57: what happens when the user inputs `"q"` here? lets write a test for that

ConwaysGameOfLife/OutputFormatter.cs
```diff
@@ -10,11 +10,12 @@ public static class OutputFormatter
         public static string FormatUniverse(Universe universe) // universe created with user input - dimensions & live cells ?? - needed??
         {
             StringBuilder stringBuilder = new StringBuilder();
-            for (var y = 0; y < universe.GridLength; y++)
+            for (var y = 0; y < universe.GridLength; y++)// TODO: problem here?? - should width be first loop?
```
kedoodle, line 13: we have tests that validate that this function is working as intended. it's easy to get the x,y mixed up but we can fall back on our existing tests to realise that this function is working.

ConwaysGameOfLife/OutputFormatter.cs
```diff
@@ -10,11 +10,12 @@ public static class OutputFormatter
...
         {
             StringBuilder stringBuilder = new StringBuilder();
-            for (var y = 0; y < universe.GridLength; y++)
+            for (var y = 0; y < universe.GridLength; y++)// TODO: problem here?? - should width be first loop?
             {
                 for (var x = 0; x < universe.GridWidth; x++)
                 {
-                    var state = universe.GetCellAtLocation(x, y).CellState;
+                    var location = new Location(x, y); 
+                    var state = universe.GetCellAtLocation(location).CellState;
```
kedoodle, line 18: nice 👍

ConwaysGameOfLife/Universe.cs
```diff
@@ -74,7 +74,7 @@ public int CountLiveNeighbours(Cell cell)
             var neighbourLocations = GetCellNeighbourLocations(cell);
             foreach (var neighbour in neighbourLocations)
             {
-                var state = GetCellAtLocation(neighbour.X, neighbour.Y).CellState;
+                var state = GetCellAtLocation(neighbour).CellState;
```
kedoodle, line 77:
```suggestion
                var cell = GetCellAtLocation(neighbour);
                var state = cell.CellState;
```

as you had done above

kedoodle (approved):
looking good. lets aim to continue to replace the existing code/tests in the controller with code/tests for the run function

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -18,128 +18,150 @@ public GameControllerShould()
...
-        public void ReturnUniverseString_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput)
+        [InlineData(new string[] { "0,3", "3,0" }, "3,3", "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        [InlineData(new string[] { "0;3", "3-0", "4 4" }, "3,3", "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        [InlineData(new string[] { ",," }, "9,9", "💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n")]
+        [InlineData(new string[] { "3,3," }, "9,9", "💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n💀💀💀💀💀💀💀💀💀\n")]
+        [InlineData(new string[] { "3,3,3", "3;3;" }, "3,3", "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        [InlineData(new string[] { "a,b" }, "3,3", "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        [InlineData(new string[] { "-1,4", "4,-1" }, "3,3", "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        [InlineData(new string[] { "10,0", "0,10" }, "3,3", "💀💀💀\n💀💀💀\n💀💀💀\n")]
+        public void DisplayUniverse_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput, string expected)
```
indianbread, line 139:
```suggestion
        public void CreateUniverse_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput, string expected)

```

indianbread: What is tested here is that the universe was created successfully from the valid input at the end, and the displaying of the universe is more of how it's tested

ConwaysGameOfLife/GameController.cs
indianbread: At which point does the user have the option to quit the application - is it immediately after the welcome message is displayed and before requesting the dimensions, or only after the set up is finished? (up to you - but this may affect the structure of this method)

ConwaysGameOfLife/GameController.cs
```diff
@@ -42,19 +42,57 @@ public GameController(IInput input, IOutput output)
...
             // PopulateUniverseWithLiveCells();
             // DisplayUniverse();
             // generator checks all cells if change of state required & creates next universe
             // loop last step until user presses 'q' to quit or all cells are dead
         }
 
-        public void CreateInitialUniverse()
+        private bool UserEndsGame(string input)
+        {
+            if (input == "q")
```
indianbread, line 85:
```suggestion
           return input == "q"

```

ConwaysGameOfLife/GameController.cs
```diff
@@ -42,19 +42,57 @@ public GameController(IInput input, IOutput output)
...
+            {
+                return true;
+            }
+            return false;
+        }
+
+        public void CreateInitialUniverse(string input)
```
indianbread, line 92: why pass in an input argument if it will get overridden anyway?

ConwaysGameOfLife/GameController.cs
```diff
@@ -103,11 +138,11 @@ public string CreateValidLocationString()
             return input;
         }
 
-        public Cell SetLiveCellLocation(Location location)
+        public Universe ReturnUniverseAfterSettingLiveCellLocation(Location location)
```
indianbread, line 141: this method is doing 2 things - changing the cell state and returning a universe. recommend to separate these concerns & rename the method

indianbread (approved):

ConwaysGameOfLife.Tests/GameControllerShould.cs
lindajohnstone: What do I do so that I don't keep making this same mistake?

PR #19, opened 2021-06-28
21 06 28

lindajohnstone:


ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -120,9 +69,9 @@ public void DisplayUniverse_GivenValidUniverseString(string input, string expect
         {
             _input.GetReadLine(input);
             _input.GetReadLine("q");
-            
+
             _controller.Run();
-            var actual = _output.GetLastWriteLine();
+            var actual = _output.OutputList[^4];
```
indianbread, line 74: As an outsider reading your code, I don't want to have to go into your function implementation to determine why it's checking the 4th last element of your array. I suggest wrapping this in a function and giving it a clear name so the intent of this is clear. One way to do this is to add an additional function in your stub output - something like DisplayUniverse which will do this

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -136,21 +85,21 @@ public void DisplayUniverse_GivenValidUniverseString(string input, string expect
...
-        public void DisplayUniverse_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput, string expected)
+        public void CreateUniverse_GivenInvalidInputsFollowedByValidInput(IEnumerable<string> invalidInputs, string validInput, string expected)
         {
             _input.GetReadLine(invalidInputs);
             _input.GetReadLine(validInput);
             _input.GetReadLine("q");
 
             _controller.Run();
-            var actual = _output.GetLastWriteLine();
+            var actual = _output.OutputList[^4];
```
indianbread, line 95: see previous comment

indianbread (approved):
Good progress today. Can you spot some opportunities for refactoring in your Run function now that it's further along?

ConwaysGameOfLife.Tests/GameControllerShould.cs
kedoodle: It varies which output is expected to be a formatted `Universe` . Not entirely sure how best to approach this. Could just give `actual` a more meaningful name e.g. `actualFormattedUniverse`.

PR #20, opened 2021-06-29
21 06 29

lindajohnstone:


ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -1,10 +1,75 @@
...
+        [Theory]
+        [MemberData(nameof(UniverseMemberData.GetUniverseFromDataGenerator), MemberType = typeof(UniverseMemberData))]
+        //[MemberData(nameof(UniverseMemberData.UniversesFromStrings), MemberType = typeof(UniverseMemberData))]
+        public void CreateNewUniverse_FromUniverseWithLiveCells(string[] rows, string[] rowsNextUniverse)
+        {
+            var initData = String.Join(Environment.NewLine, rows);
+            var universe = UniverseHelper.InitializeUniverse(initData);
+            var generator = new Generator(universe);
+            var initDataNextUniverse = String.Join(Environment.NewLine, rowsNextUniverse);
+            var nextUniverse = UniverseHelper.InitializeUniverse(initDataNextUniverse);
```
indianbread, line 67:
```suggestion
            var expected = UniverseHelper.InitializeUniverse(initDataNextUniverse);

```

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -1,10 +1,75 @@
...
+            var expected = nextUniverse;
```
indianbread, line 68:
```suggestion
```

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -1,10 +1,75 @@
...
 namespace ConwaysGameOfLife.Tests
 {
     public class GeneratorShould
     {
-        
+        [Theory]
+        [InlineData("XXX", "XXX", "XXX")]
+        [InlineData("XX", "XX", "XX")]
+        [InlineData("XXXX", "XXXX", "XXXX")]
+        public void ReturnNewListOfCells_FromOldUniverse(params string[] rows)
```
indianbread, line 13: Is this test required? When testing the `GenerateNewUniverse` function what matters is that when given an old universe the correct new universe is generated. When checking that the universes are equal in the later tests the universe helper function `UniversesAreEqual` is checking the list of cells anyway.

ConwaysGameOfLife.Tests/UniverseHelper.cs
```diff
@@ -34,6 +34,23 @@ public static bool UniversesAreEqual(Universe universe1, Universe universe2)
             return true;
         }
 
+        public static bool ListsOfCellsAreEqual(List<Cell> cells1, List<Cell> cells2)
```
indianbread, line 37: probably not required - see my comment above

ConwaysGameOfLife.Tests/UniverseMemberData.cs
```diff
@@ -17,5 +18,103 @@ public static class UniverseMemberData
                 new Cell(CellState.Dead, 2, 1),
                 new Cell(CellState.Dead, 2, 2)
             };
+
+        public static TheoryData<string[]> UniversesFromStrings =
+            new TheoryData<string[]>
```
indianbread, line 23:
```suggestion
            new TheoryData<object[]> 

```
either that or an array of string arrays

ConwaysGameOfLife.Tests/UniverseMemberData.cs
```diff
@@ -17,5 +18,103 @@ public static class UniverseMemberData
                 new Cell(CellState.Dead, 2, 1),
                 new Cell(CellState.Dead, 2, 2)
             };
+
+        public static TheoryData<string[]> UniversesFromStrings =
+            new TheoryData<string[]> 
+        // TODO: fails. 
+        // error: System.InvalidOperationException : The test method expected 2 parameter values, but 1 parameter value was provided.
+            {
+                new string[]
```
indianbread, line 27:
```suggestion
                new string[]
{
```

ConwaysGameOfLife.Tests/UniverseMemberData.cs
```diff
@@ -17,5 +18,103 @@ public static class UniverseMemberData
...
+                    "OXX",
+                    "OXX",
+                    "OXX"
+                },
+                new string[]
+                {
+                    "OOO",
+                    "OOO",
+                    "OOO"
+                }
```
indianbread, line 38:
```suggestion
                }
           }

```

indianbread: a collection (e.g. array) of string arrays

ConwaysGameOfLife/Generator.cs
```diff
@@ -28,13 +32,39 @@ public Generator()
...
+            var oldCells = _oldUniverse.Cells;
+            var newCells = new List<Cell>();
+            foreach (var cell in oldCells)
+            {
+                GenerateCells(newCells, cell);
+            }
+            return new Universe(_oldUniverse.GridWidth, _oldUniverse.GridLength, newCells);
+        }
+
+        private void GenerateCells(List<Cell> cells, Cell cell)
```
indianbread, line 46: this function can return a list of cells as it's what the function name is suggesting.

ConwaysGameOfLife/Generator.cs
```diff
@@ -28,13 +32,39 @@ public Generator()
...
         }
-        public Universe GenerateNewUniverse(Universe universe)
+
+        private Cell GenerateNewCell(Cell cell)
         {
-            throw new NotImplementedException();
+            var numberOfLiveNeighbours = _oldUniverse.CountLiveNeighbours(cell);
+            var state = cell.CellState;
+            var newCell = new Cell(state, cell.Location.X, cell.Location.Y);
+            ShouldSwitchCellState(numberOfLiveNeighbours, state, newCell);
```
indianbread, line 57: Does calling `ShouldSwitchCellState` here make sense? What if the cell doesn't need to switch cell state?
Method name indicates it should return a boolean.
If it should switch cell state, set the new state in this function.

ConwaysGameOfLife/Generator.cs
```diff
@@ -28,13 +32,39 @@ public Generator()
 
         */
 
-        public bool ShouldSwitchCellState()
+        public Universe GenerateNewUniverse()
+        {
+            var oldCells = _oldUniverse.Cells;
+            var newCells = new List<Cell>();
```
indianbread, line 38: move all code pertaining to the generation of the new list of cells to GenerateCells, then assign the return value to the newCells variable:

`var newCells = GenerateCells();`

indianbread (approved):

ConwaysGameOfLife.Tests/GeneratorShould.cs
kedoodle: I recall you had a doc somewhere outlining roughly what tests might be in each class. It may be worthwhile updating that doc and thinking about what tests are and aren't necessary.

Ideally, your tests should be validating some kind of program behaviour rather than the specific implementation details i.e. in this test, we should really be testing that the `Universe` being generated by the `Generator` is what we're expecting. An implementation detail of that is that the `Cell` list inside the `Universe` is equal to what we're expecting (and one step further down the chain, the `Location` and `CellState` inside the cell is equal to what we're expecting etc.).

ConwaysGameOfLife.Tests/GeneratorShould.cs
```diff
@@ -1,10 +1,75 @@
...
+        {
+            var initData = String.Join(Environment.NewLine, rows);
+            var universe = UniverseHelper.InitializeUniverse(initData);
```
kedoodle, line 16: Would it help to make `InitializeUniverse` take `params string[] rows` instead of `string initData`? To save doing the `Join` each time it's being used

ConwaysGameOfLife.Tests/UniverseHelper.cs
kedoodle: side note: should equality be reliant on the ordering of the items in the list? compare this to your implementation in `UniversesAreEqual`

ConwaysGameOfLife/Generator.cs
```diff
@@ -28,13 +32,39 @@ public Generator()
...
+        }
+
+        private void ShouldSwitchCellState(int numberOfLiveNeighbours, CellState state, Cell newCell)
+        {
+            var shouldNewCellHaveDifferentCellState = _rules.Any((rule) => rule.ShouldSwitchCellState(numberOfLiveNeighbours, state));
+            if (shouldNewCellHaveDifferentCellState)
+            {
+                newCell.SwitchCellState();
+            }
         }
```
kedoodle, line 68: please read into command query separation. it's holding you back and leading you to produce functions like this that have "side effects" which can cause confusion to your future self (or whoever the maintenance programmer/reader is)

```suggestion
        private bool ShouldSwitchCellState(int numberOfLiveNeighbours, CellState state)
        {
            return _rules.Any((rule) => rule.ShouldSwitchCellState(numberOfLiveNeighbours, state));
        }
```

ConwaysGameOfLife/Generator.cs
kedoodle:
```suggestion
            if (ShouldSwitchCellState(numberOfLiveNeighbours, state)) newCell.SwitchCellState();
```

kedoodle (approved):
good overall. nice to see the active refactoring.

PR #22, opened 2021-07-01
21 07 01

lindajohnstone:
Sorry for the late p/r. 

@Kedoodle, @indianbread & I spent time this afternoon working out how to test ReadKey. So far, all changes have been unsuccessful. Not spent any time on it since the meeting due to other meetings. Will look at it again tomorrow.


indianbread (approved):
If still no success with testing the console key available tomorrow it's probably best to make a new branch and check the input using strings

ConwaysGameOfLife.Tests/GameControllerShould.cs
```diff
@@ -23,8 +23,10 @@ public GameControllerShould()
         public void EndRunImmediately_GivenQInput()
         {
             var expectedOutput = "Game of Life has ended.";
+            
             var input = "q";
-            _input.GetReadLine(input);
+            _input.GetReadLine("3,3");
+            _input.GetReadKey(input);
```
kedoodle, line 29: This change makes it so the test is checking if you can quit after a `Universe` has been created and the game is now prompting for a `Location` i.e. the test is no longer reflecting what the name implies. It's more or less doing the same as the `EndGame_GivenQInputDuringLocationInput` test in terms of behaviour

kedoodle (approved):
you might find it hard to test the `Console.KeyAvailable` strategy without mocking the `Console` i.e. one which you can assign whether the boolean returns true or false at each given time

PR #23, opened 2021-07-02
21 07 02

lindajohnstone:
Two issues today:
The game is now set to run in the console. However, it never gets past the setup stage - input p is seen as invalid input. I tried many ways to fix this (& **tried** to commit each time).
The other issue is I tried to test the game ending on Universe.AreAllCells. However, the only way to ensure the input queue is not empty is to use '_input.GetReadLine("q")' which ends the game anyway & a second universe is not created.

### DIFF
--- a/ConwaysGameOfLife/GameController.cs
+++ b/ConwaysGameOfLife/GameController.cs
@@ -53,7 +53,10 @@ namespace ConwaysGameOfLife
                         _output.WriteLine(Messages.GameEnd);
                         return;
                     }
-                    
+                    // if (UserPressesPToPlay(input)) // ends game
+                    // {
+                    //     return;
+                    // }
 
                     var isValidUniverse = Validator.IsValidUniverse(input);
                     while (!isValidUniverse)
@@ -66,6 +69,10 @@ namespace ConwaysGameOfLife
                             _output.WriteLine(Messages.GameEnd);
                             return;
                         }
+                        // if (UserPressesPToPlay(input))
+                        // {
+                        //     return;
+                        // }
                         isValidUniverse = Validator.IsValidUniverse(input);
                     }
                     _universe = InputParser.ParseUniverse(input);
@@ -77,13 +84,16 @@ namespace ConwaysGameOfLife
                         _output.WriteLine($"or {Messages.Play}");
 
                         input = _input.ReadLine();
-
+            
                         if (UserEndsGame(input))
                         {
                             _output.WriteLine(Messages.GameEnd);
                             return;
                         }
-
+                        // if (UserPressesPToPlay(input))
+                        // {
+                        //     return;
+                        // }
                         var isValidLocation = Validator.IsValidLocation(input, _universe.GridWidth, _universe.GridLength);
                         while (!isValidLocation)
                         {
@@ -96,6 +106,12 @@ namespace ConwaysGameOfLife
                                 _output.WriteLine(Messages.GameEnd);
                                 return;
                             }
+
+                            // if (UserPressesPToPlay(input))
+                            // {
+                            //     return;
+                            // }
+                            
                             isValidLocation = Validator.IsValidLocation(input, _universe.GridWidth, _universe.GridLength);
                         }
                         var location = InputParser.ParseLocation(input);
@@ -103,7 +119,7 @@ namespace ConwaysGameOfLife
                         SetLiveCellLocation(location);
                         DisplayUniverse();
                     }
-                    while (input != "p");
+                    while (!UserPressesPToPlay(input));
                     Play();
             //     }
             // }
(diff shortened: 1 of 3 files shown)


indianbread (approved):
Your run function and hence game loop is quite big and confusing with the commented out code here and there. Instead of commenting out what is not working and jumping ahead, I strongly recommend you start from the outside in (i.e. from the outermost loop where q = quit) and make sure one functionality works correctly first before you move on to another - even if this means removing everything inside the loop and slowly adding them back in. 
Use the troubleshooting steps that we performed together previously:
1) If the input queue is empty, use the debugger to step through the code and see what is being enqueued/dequeued at which line(s)
2) If a function call is not producing the desired output, use the debugger and step through each line. Do the variables have the expected value at each stage of the code execution? Is the code executing in the expected order, or jumping to an unexpected place?

ConwaysGameOfLife/GameController.cs
```diff
@@ -76,51 +84,61 @@ public void Run()
...
         {
-            //var generator = new Generator();
-
-            // loop until all cells are dead or user enters q
-            var generator = new Generator(_universe);
-            generator.GenerateNewUniverse();
+            while(!_universe.AreAllCellsDead())
+            {
+                var generator = new Generator(_universe);
+                generator.GenerateNewUniverse();
```
kedoodle, line 134: Not saving the generated `Universe` in any way here
```suggestion
                _universe = generator.GenerateNewUniverse();
```

ConwaysGameOfLife/GameController.cs
```diff
@@ -76,51 +84,61 @@ public void Run()
...
+                            
                             isValidLocation = Validator.IsValidLocation(input, _universe.GridWidth, _universe.GridLength);
                         }
                         var location = InputParser.ParseLocation(input);
 
                         SetLiveCellLocation(location);
                         DisplayUniverse();
                     }
-                    while (input != "p");
+                    while (!UserPressesPToPlay(input));
```
kedoodle, line 122:
```suggestion
                    while (!UserPressesPToPlay(input));
```
another potential infinite loop here as you're never updating `input` i.e. if you get into this loop because `input` is not `"p"`, you'll never get out

kedoodle (approved):
It might be frustrating knowing that you're so close to finishing this kata, but can't quite get this last `GameController.Run()` function working as you'd like. The good thing is you've actually got all of the pieces you need to get it working. It's just about putting it all together now. I'd suggest taking it back to basics and thinking logically about what you want to happen, step by step e.g.

1. User starts the program
2. Program welcomes user to game
3. Game prompts user for `Universe` input
a. User responds `"q"` -> exit game
b. User responds any invalid `Universe` string -> move to 3
c. User responds any valid `Universe` string -> move to 4
4. Game prompts user for `Location` input
a. User responds `"q"` -> exit game
b. User responds any invalid `Location` string -> move to 4
c. User responds any valid `Location` string -> switch state of cell in location, move to 4
d. User responds `"p"` -> move to 5
5. If game ended i.e. all cells dead -> exit game
6. Generator created and new `Universe` shown
7. If game ended i.e. all cells dead -> exit game
8. User prompted for input
a. User responds `"q"` -> exit game
b. User responds `"p"` -> move to 6
c. User responds anything else -> move to 8

It sounds silly, but sometimes this is all you need. Pseudocode.


As for more practical, actual code, think about why you might have a `do while` loop. Usually, it's because you want to go through the loop at least one iteration before checking the condition.

In your case, it might be because you want to get the user's input and/or check the game isn't over due to the whole `Universe` being dead. This is just an example. You may find that you don't want a huge `do while` loop at all. It might be easier without one...

 The game can't move backwards in "stage/state" if you think of it as a few distinct stages/states:
- There's a greeting stage; this one is pretty self explanatory and you move on from it without any user input.
- A `Universe` initialisation stage where the dimensions of the `Universe` are set. The user can choose to quit during this stage. The user moves on from this stage by providing a valid `Universe` string input
- A `Location` initialisation stage where the cell states of the `Universe` are set. The user can choose to quit during this stage. The user can also choose to move on from this stage by providing a `"p"` input
- A "playing" stage where the game is actually played i.e. new `Universe` are generated based on the rules and the current `Universe`. The user can choose to quit during this stage. The game can also be ended without the user explicitly entering `"q"` by way of a completely dead `Universe` being generated
At no point can the game move backwards in stages. This plays to your advantage as you can "solve"/code the stages one by one and actually try to play the game through as you go (it could just end abruptly until you've coded the next stage).

Now, back to the loops. Do the stages make it clearer where you may/may not need a loop? Clearly, there's no need to loop when we're not asking for user input i.e. in the welcome stage. But what about the other stages - how many loop(s) do you think we need? Go back to the step-by-step pseudocode and see if you can identify where the loops might be needed.

PR #24, opened 2021-07-05
21 07 05

lindajohnstone:
I need to note down:
Reflections 
- what went well
- learnings
Challenges
Extensions

Is this all?


indianbread (approved):
hooray for wrapping up game of life. I recommend keeping a record (can use any tool that you like e.g. one note, apple notes, notion) of each kata you complete. Document:
* your technical & non technical learnings
* the challenges and issues you experienced
* lessons learnt - technical & non technical
* what you would do differently next time - could be technical or non technical
* Extensions

kedoodle (approved):
Nice one. Yeah I think it would be a good exercise to take some time to reflect on the process through this kata:
- learnings
- what didn't go so well e.g. what could have gone differently to avoid similar frustrations next time
- what went well
- key takeaways (can be learnings)
- overall thoughts

